PR #105, opened 2015-02-21
Fix prompt precmd spacing.

mafredri:
Only add spaces if there is content.


sindresorhus:
Thanks :)

mafredri:
No problem, glad I could help, really liking the theme :)

Issue #81, opened 2014-09-02
Pure creates lots of git-fetch and git-upload-pack in ssh repositories


fedemp:
When I enter a repository whose origin url uses ssh (i.e. git://gitlab.foo.com/repo/repo.git), pure launches one after another of `git fetch` and `ssh` processes.

````
git fetch
ssh git@gitlab.foo.com git-upload-pack 'repo/repo.git'
````

All proccesses are listed as `TN`.

sindresorhus:
How did you see the processes?

fedemp:
Is this what you are asking for?

````
› ps ax | grep git
27905 pts/2    TN     0:00 git fetch
27906 pts/2    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
27923 pts/2    TN     0:00 git fetch
27924 pts/2    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
27966 pts/2    TN     0:00 git fetch
27967 pts/2    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
27983 pts/2    TN     0:00 git fetch
27984 pts/2    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28335 pts/2    TN     0:00 git fetch
28336 pts/2    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28409 pts/1    TN     0:00 git fetch
28410 pts/1    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28748 pts/5    TN     0:00 git fetch
28749 pts/5    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28764 pts/5    TN     0:00 git fetch
28765 pts/5    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28782 pts/5    TN     0:00 git fetch
28783 pts/5    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28799 pts/5    TN     0:00 git fetch
28800 pts/5    TN     0:00 ssh git@gitlab.foobar.com git-upload-pack 'owner/super-project-ui.git'
28829 pts/5    TN     0:00 git fetch
[...]

PR #106, opened 2015-02-21
Do not draw unpushed/unpulled arrows for stale precmd sessions.

mafredri:
Sometimes arrows are drawn multiple times due to stale precmd's still running which contain the wrong length for the prompt line (i.e. `prompt_pure_string_length $prompt_pure_preprompt` contains stale info). This change will try to stop a forked precmd from executing if another one has been started after it.

There are multiple checks in place so as to stop execution as soon as possible.

This implementation uses a temp file, there are probably better alternatives to keep state than that, but... The best solution would be to allow only one of these forks to run at a time / git repo, but that would require much more logic.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -64,6 +64,10 @@ prompt_pure_string_length() {
 	echo ${#${(S%%)1//(\%([KF1]|)\{*\}|\%[Bbkf])}}
 }
 
+prompt_pure_precmd_valid() {
+	[[ "$1" == "$(<$prompt_pure_last_cmd_tmp)" ]] || return 1
+}
+
 prompt_pure_precmd() {
 	# shows the full path in the title
 	print -Pn '\e]0;%~\a'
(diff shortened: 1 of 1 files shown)


sindresorhus:
I have been noticing this sometimes too, but never have been able to consistently reproduce it. Thanks for looking into it.

Wouldn't using an environment variable make more sense than a temp file?

mafredri:
I looked into using an environment variable and hit a dead end pretty early. Problem is that since the process is forked by `&!` there is no way to access the updated environment variables. So even if you update an environment variable later on, it will not transfer to the process running.

I'm open to refactoring this with a better method that does not require a temp-file. But I'm sure my ZSH-foo is currently too limited to figure anything out regarding that :).

mafredri:
I'm investigating using coproc to fix this, should be a pretty elegant solution.

mafredri:
Ok, I went completely back to the drawing board in regards to this feature. Rewrite most of the git fetch logic.

I'm open to discuss the changes and alternative implementations. There are no more temporary files, I do however use a kill -SIGINFO signal from the disowned process to notify the shell that the git fetch task has been completed.

This also fixes https://github.com/sindresorhus/pure/issues/81, at least partially. There can still be multiple `git fetch`es running if you are quick with changing directories.

Cheers,
Mathias

Issue #102, opened 2015-01-18
Pure is very slow when using in cdnjs' repo


nesk:
I just forked [cdnjs' repo](https://github.com/cdnjs/cdnjs) and each command takes ~25s to execute when using pure. I disabled remote checking (`PURE_GIT_PULL`) and untracked files checking (`PURE_GIT_UNTRACKED_DIRTY `) but it doesn't bring any improvements.

Am I missing something?

sindresorhus:
Yeah, that's painful. git get's really slow on large repos. That option only enables a faster mode, it doesn't disable dirty checking all together.

I've been thinking and instead of introducing yet another option I could make the dirty-checking timeout after some predefined time, like 1 second. That way it would never annoyingly block for a long time, but also don't require an option.

Nesk:
I'm totally for this timeout, but the user should be aware when it has been reached (for example, with a yellow prompt character).

It seems like there is a “not-so-complex” solution [on this page](https://digitalelf.net/2013/08/timeout-in-a-shell-script/), however I doesn't seem to work very well with zsh, unlike the author is saying.

sindresorhus:
> I'm totally for this timeout, but the user should be aware when it has been reached (for example, with a yellow prompt character).

Agreed.

paulirish:
Dirty checking is about 2.6x faster with this:

```sh
git diff --no-ext-diff --quiet --exit-code
```

ref: https://github.com/git/git/blob/355d4e1739d1e88561ac04d2a2423c2aa93693d9/contrib/completion/git-prompt.sh#L475

arthurvr:
Thanks, @paulirish! Awesomesauce! Makes sense when working with this kind of repos (chromium repo is even worse). I'm +1 on a timeout.

paulirish:
in my case i'm going to blacklist chromium & blink for dirty checking. less
robust but easier to manage.​

sindresorhus:
This should be a faster in 1.0.0:
https://github.com/sindresorhus/pure/releases/tag/1.0.0

If a git status takes longer than 2 second, it will rely on cached result for 30 minutes. This is configurable: https://github.com/sindresorhus/pure#pure_git_delay_dirty_check

Though I opened a new issue about improving the dirty check: https://github.com/sindresorhus/pure/issues/115

-

// @mafredri

PR #108, opened 2015-02-28
Refactor async logic and allow for async git status.

mafredri:
This pull request replaces my previous PR (https://github.com/sindresorhus/pure/pull/106) and fixes issues https://github.com/sindresorhus/pure/issues/81 and https://github.com/sindresorhus/pure/issues/102. 

Discussions welcome.

- Pure now uses a background worker for git fetch and git status
- Only one git fetch/status can run at a time
- All worker jobs are eliminated on directory change (consider passing
only git working tree?)
- If a git status takes longer than 2 second, rely on cached result
for 30 minutes (1800 seconds)
- Disowned process no longer takes care of rendering the git up/down
arrows. It is done by the parent process
- Entire preprompt can be redrawn to allow for updating git status,
etc. Works well in both smaller and larger terminals.
- Arrows can be drawn immediately if the data is available

Some ideas:
* Extend the periodical check to perform new checks even when no new action has been take on the terminal. So even a background terminal could inform you that there are changes to pull, etc.
* Make git status execution time-limit configurable (instead of hard-coded 2 seconds)
* Now worker checks are performed 1s after execution and periodically every 15 seconds. Maybe check for worker results more often until children have performed results. Could also increase periodic time to e.g. 1min and make it configurable.

### DIFF
--- a/arch/PKGBUILD
+++ b/arch/PKGBUILD
@@ -21,4 +21,6 @@ package() {
   cd $srcdir/$pkgname
   install -Dm644 pure.zsh \
     "$pkgdir/usr/share/zsh/functions/Prompts/prompt_pure_setup"
+  install -Dm644 async.zsh \
+    "$pkgdir/usr/share/zsh/functions/async"
 }
--- a/async.plugin.zsh
+++ b/async.plugin.zsh
@@ -1,0 +1,1 @@
+async.zsh
--- a/async.zsh
+++ b/async.zsh
@@ -1,0 +1,215 @@
+#!/usr/bin/env zsh
+
+#
+# zsh-async
+#
+# version: 0.1.0
+# author: Mathias Fredriksson
+# url: https://github.com/mafredri/zsh-async
+#
+
+# Wrapper for jobs executed by the async worker, gives output in parseable format with execution time
+_async_job() {
+	local out
+	# store the job identifier
+	local job=$1
+	# reset the identifier
+	1=""
+	# store start time
+	local start=$EPOCHREALTIME
+
+	# run the command
+	out=$($job $@ 2>&1)
+	local ret=$?
+
+	# Grab mutex lock
+	read -ep >/dev/null
+
+	# return output
+	print -r -N -n -- $job $ret "$out" $(( $EPOCHREALTIME - $start ))$'\0'
+
+	# Unlock mutex
+	print -p "t"
+}
+
+# The background worker manages all tasks and runs them without interfering with other processes
+_async_worker() {
+	local -A storage
+	local unique=0
+
+	# Process option parameters passed to worker
+	while getopts "np:u" opt; do
+		case "$opt" in
+		# Use SIGWINCH since many others seem to cause zsh to freeze, e.g. ALRM, INFO, etc.
+		n) trap 'kill -WINCH $ASYNC_WORKER_PARENT_PID' CHLD;;
+		p) ASYNC_WORKER_PARENT_PID=$OPTARG;;
+		u) unique=1;;
+		esac
+	done
+
+	# Create a mutex for writing to the terminal through coproc
+	coproc cat
+	# Insert token into coproc
+	print -p "t"
+
+	while read -r cmd; do
+		# Separate on spaces into an array
+		cmd=(${=cmd})
+		local job=$cmd[1]
+
+		# Check for non-job commands sent to worker
+		case "$job" in
+		_killjobs)
+			kill ${${(v)jobstates##*:*:}%=*} &>/dev/null
+			continue
+			;;
+		esac
+
+		# If worker should perform unique jobs
+		if ((unique)); then
+			# Check if a previous job is still running, if yes, let it finnish
+			for pid in ${${(v)jobstates##*:*:}%=*}; do
+				if [[ "${storage[$job]}" == "$pid" ]]; then
+					continue 2
+				fi
+			done
+		fi
+
+		# run task in background
+		_async_job $cmd &
+		# store pid because zsh job manager is extremely unflexible (show jobname as non-unique '$job')...
+		storage[$job]=$!
+	done
+}
+
+#
+#  Get results from finnished jobs and pass it to the to callback function. This is the only way to reliably return the
+#  job name, return code, output and execution time and with minimal effort.
+#
+# usage:
+# 	async_process_results <worker_name> <callback_function>
+#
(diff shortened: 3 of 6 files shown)


mafredri:
I made my async implementation into a library: https://github.com/mafredri/zsh-async

I felt that I cluttered down your theme pretty radically with all my code changes. Thoughts about importing a "library" instead?

sindresorhus:
Awesome! :)

> consider passing only git working tree?

Yes

> Extend the periodical check to perform new checks even when no new action has been take on the terminal. So even a background terminal could inform you that there are changes to pull, etc.

How would that work? If there were any changes, would my prompt suddenly get the arrow interactively?

> Make git status execution time-limit configurable (instead of hard-coded 2 seconds)

I'd prefer not to unless there's a good reason.

> If a git status takes longer than 2 second, rely on cached result for 30 minutes (1800 seconds)

Maybe we should use a separate color on the mark so the user is aware?

> Now worker checks are performed 1s after execution and periodically every 15 seconds. Maybe check for worker results more often until children have performed results. Could also increase periodic time to e.g. 1min and make it configurable.

Can you elaborate, is the "worker checks" doing "git fetch" every 15 seconds or do you mean check for result from the worker?

mafredri:
>> consider passing only git working tree?

> Yes

Agreed, this is the best course of action here, no point in stopping a process unless we are moving out of the git working tree. Will fix.

> How would that work? If there were any changes, would my prompt suddenly get the arrow interactively?

Precisely! Say you have two open terminals for the same git directory, both old, perform git fetch on one (assuming changes locally / upstream) the other terminal would be updated with arrows during the next periodical check.

Currently the periodic check is run every 15 minutes, but it only checks for the tasks run by that terminal. This could go two ways (although I'm not too sure how I feel about the terminal performing tasks in the background without intervention):

0. Only run periodic check (e.g. every second) when there are tasks active, then stop 
0. Combine 1. and run additional periodic check e.g. every 1 minute which could also launch new git status / check for local / remote commits

> I'd prefer not to unless there's a good reason.

I can appreciate that, it was just a thought since 2 seconds was something I kind of pulled out of the hat. Felt that maybe 1 second is being a bit too aggressive (i.e. not that stressful) and more than 2 then there might be an actual cost somewhere.

> Maybe we should use a separate color on the mark so the user is aware?

That's a great idea! I was only thinking about possibly indicating that `git status` is running, and indicating that it's "cached" hadn't crossed my mind :)

> Can you elaborate, is the "worker checks" doing "git fetch" every 15 seconds or do you mean check for result from the worker?

Yes, the worker check is only a check for results from a task that has been launched earlier (e.g. git status, git fetch). So if a task has not been launched by other means, the worker check will return empty handed. I guess this point kind of intersected with the earlier one about periodic checks, hope I brought some clarity to the table.

> I would love to use a lib like yours instead. How would we do that though?

Are you asking about how we would include my lib in this project? If so, I'm not quite sure. I guess if I pushed it to npm, then that would allow pure to require it, but then that would not allow for easy independent installation. Also, having it as a git submodule would be one possibility, but that requires the additional submodule pulling / etc. Third option I can think of would be maybe to include it in this project as a separate file, updates would need to be imported retroactively.

Got any other ideas? In my lib I've already fixed some potential problems which exist in this PRs implementation, e.g. race-conditions for tasks that finish nearly at the same time, etc. So would be nice to rely on it instead of a custom one here :).

sindresorhus:
Thanks for clarifications. Sounds good to me.

> Third option I can think of would be maybe to include it in this project as a separate file, updates would need to be imported retroactively.

Let's go with this.

mafredri:
Alright, that's great!

I've done most of the refactor for this already, but it still needs some fixing here and there. One thing I'm still trying to work out is something that would allow to get rid of the polling (scheduling) and that is sending a kill signal to the parent process. It works beautifully, and feels much more interactive than polling, but there's an edge case ZSH completely freezes if you send multiple kill signals at the same time / during user interaction. (I.e. holding enter in the terminal can eventually crash it). I'm thinking of rate-limiting kill-signals on the async-library side, but it's taking some time to crack :)...

sindresorhus:
Oh, yeah, some kind of signal would be much better than polling. Awesome work on this btw :)

mafredri:
Thanks, it's been a very painful process :)! Obviously something zsh wasn't meant to do :D..

I think I finally found a solution, or workaround rather, to the kill signal leading to zsh freezing up. I'm now sending a SIGWINCH (Window size change) which does not, for some magical reason, seem to be plagued by the lockups.

I'm not fully satisfied with how the async library needs to be enabled at the moment (updated the readme also) but it's the only way I can think of, thoughts?

mafredri:
Looks like me rebasing onto master moved the commits to the end, sorry about that.

Anyway, with my last two commits I thought it might be smart (since it's now possible) to move the pull/push arrows beside the git branch. Thoughts?

New arrow placement

mafredri:
With my last commits, I believe this branch is quite ready. I've been using it as my daily driver since the beginning, so most of issues should've been ironed out. There is an indication of "cached" git status now as well.

Cached git status

sindresorhus:
This looks incredible! I just need to get some time to read through the code and some days to have it running on my own computer.

sindresorhus:
Ok, finally had time to extensively try this out. Sorry about the long delay... It works perfectly! :) It might be a placebo effect, but I feel the normal prompt is faster now too. Super nice work @mafredri.

@mafredri Can you do one final rebase to fix the merge conflict?

mafredri:
Glad you liked it @sindresorhus! I've updated the PR, but If you want I could still squash the commits together so that it doesn't pollute the git history quite as much.

arthurvr:
I've been using this for a week or so and it's awesome :+1: Glorious work!

sindresorhus:
Landed! :)

sindresorhus:
@mafredri Would you happen to be interested in being the maintainer of `pure`? You've done an excellent job with this PR and you seem a lot more adept at shell scripting than me. No worries if not though.

mafredri:
Thanks for the compliments @sindresorhus! I've never really taken over as a maintainer before, but sure, I'd be interested if you're looking to take some load of your hands :).

sindresorhus:
Cool. It's not hard, and you're already pretty much doing it. Just triaging issues, reviewing PRs and looking for ways to improve the projects :)

mafredri:
Alright, sounds good :+1: !

PR #110, opened 2015-03-26
Make sure git background fetch doesn't trigger gc

ehamberg:
Pass the configuration parameter `gc.auto=0` to `git fetch` so that we don't risk starting a garbage collection when background fetching.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -83,7 +83,7 @@ prompt_pure_precmd() {
 		# make sure working tree is not $HOME
 		[[ "$(command git rev-parse --show-toplevel)" != "$HOME" ]] &&
 		# check check if there is anything to pull
-		command git fetch &>/dev/null &&
+		command git -c gc.auto=0 fetch &>/dev/null &&
 		# check if there is an upstream configured for this branch
 		command git rev-parse --abbrev-ref @'{u}' &>/dev/null && {
 			local arrows=''


pure.zsh
```diff
@@ -83,7 +83,7 @@ prompt_pure_precmd() {
 		# make sure working tree is not $HOME
 		[[ "$(command git rev-parse --show-toplevel)" != "$HOME" ]] &&
 		# check check if there is anything to pull
-		command git fetch &>/dev/null &&
+		command git -c gc.auto=0 fetch &>/dev/null &&
```
sindresorhus: Can you use the long-flag instead of `-c`?

And should every `git` command have this?

@mafredri

mafredri: From what I can gather based on [this stackoverflow answer](http://stackoverflow.com/a/5137577) only git fetch should potentially initiate `git-gc`.

Personally I have not run into any reason to disable `gc` though, but I guess it can be useful for big repos, speeding up `git fetch` a bit.

For reference, first try:
```
~/Projects/cdnjs master* ⇣
❯ time git gc
...
git gc  30.35s user 4.42s system 107% cpu 32.261 total
```
second try:
```
❯ time git gc
...
git gc  5.96s user 3.43s system 50% cpu 18.766 total
```

ehamberg:
As far as I can see, there is no long version of `-c`. From `man git`:

    -c <name>=<value>
        Pass a configuration parameter to the command. […]

Also, I *think* this only needs to be added to `fetch`.

kevva:
Alright, I think this looks good. Thanks!

Issue #116, opened 2015-05-24
`prompt pure` fails in OS X Yosemite


UltCombo:
Looks like the `pure` theme is not available after `npm install --global pure-prompt`.

```
prompt -l
Currently available prompt themes:
adam1 adam2 bart bigfade clint elite2 elite fade fire off oliver pws redhat suse walters zefram
```

mafredri:
It should be available after opening a new login session of running the command `promptinit`.

UltCombo:
I have `autoload -U promptinit && promptinit` in my `.zshrc` and I've opened up a new terminal through iTerm2.

mafredri:
Are you modifying your `fpath` somewhere perhaps and removing the defaults?

```

# echo $fpath

/usr/local/share/zsh/site-functions
/usr/local/Cellar/zsh/5.0.7/share/zsh/functions
```

mafredri:
Are you modifying your `fpath` somewhere perhaps and removing the defaults?

```
# echo $fpath
/usr/local/share/zsh/site-functions /usr/local/Cellar/zsh/5.0.7/share/zsh/functions
```

UltCombo:
```
fabricio-matte% echo $fpath
/usr/share/zsh/site-functions /usr/share/zsh/5.0.5/functions
fabricio-matte% prompt -l
Currently available prompt themes:
adam1 adam2 bart bigfade clint elite2 elite fade fire off oliver pws redhat suse walters zefram
```

Perhaps the install didn't copy the necessary files?

mafredri:
Ah, now it makes sense. The difference between using zsh bundled with OSX
and the one installed through homebrew. Please
add /usr/local/share/zsh/site-functions to your fpath and you'll be good to
go (for now).

UltCombo:
That did it, thanks!
If anyone else faces this issue, add this to your `.zshrc` before `prompt pure`:
```
fpath+=("/usr/local/share/zsh/site-functions")
```

sindresorhus:
@mafredri Is there any way we can make the experience with this better?

mafredri:
@UltCombo happy to hear! Let's see if we can't make this more smooth for others as well.

@sindresorhus It's not entirely trivial since `/usr/share/zsh/site-functions` requires root permissions, by default. Could we do a post-install message, e.g. like so:
```
Please ensure that your $fpath contains "/usr/local/share/zsh/site-functions" so that pure is accessible by zsh:
[...]

PR #119, opened 2015-05-26
possible fix for #118

### LINKED ISSUE #118: antigen is not working anymore
To make it work in antigen, I had to do something like this:

```sh
PURE_PATH=~/.antigen/repos/https-COLON--SLASH--SLASH-github.com-SLASH-sindresorhus-SLASH-pure.git

antigen bundle sindresorhus/pure
antigen apply

ln -sf "$PURE_PATH/pure.zsh" "$PURE_PATH/prompt_pure_setup"

autoload -U promptinit && promptinit
prompt pure
```

Until those last updates, only the antigen bundle/apply combo was enough... 

BTW: yes, `~/.antigen/repos/https-COLON--SLASH--SLASH-github.com-SLASH-sindresorhus-SLASH-pure.git` is in my `fpath`, but seems that `prompt` was ignoring it until I linked it to `prompt_pure_setup`.

My point is: we might need to fix documentation or fix the problem. I don't know what is the problem, but I will be glad to help testing!

caarlos0:
Don't know if this is right or if it will break something else, just a proposal..

Closes #118 

Refs #116 (probably closes it too)

### DIFF
--- a/prompt_pure_setup
+++ b/prompt_pure_setup
@@ -1,0 +1,1 @@
+pure.zsh


mafredri:
Thanks for the PR, but I don't think this is the right solution to the problem. I don't even quite understand why it would work :/.

caarlos0:
@mafredri yeah, me neither.. :/

mafredri:
Closing this now as readme has been updated with proper instructions for antigen.

Issue #121, opened 2015-05-27
It's not working with Prezto


jitendravyas:
I have set `style ':prezto:module:prompt' theme 'pure'`

and After I quit and restarted my iterm2 and getting this

![1 zsh 2015-05-27 09-28-37](https://cloud.githubusercontent.com/assets/77309/7828573/d210aabc-0452-11e5-87a6-63405c163313.png)

mafredri:
What happens if you type just `prompt pure`? Anyway, I'll have a look at prezto integration later today and see what I can do.

jitendravyas:
If i run `prompt pure` it activates `pure` but if I quit and start iterm again I get the same thing again (as above screenshot)

mafredri:
Hmm, you did write that you have set `style ':prezto:module:prompt' theme 'pure'`. It should be `zstyle ':prezto:module:prompt' theme 'pure'`.

jitendravyas:
it was my mistake to type here. it's `zstyle` only in an actual file. By the way, I changed `pure` to `sorin` now and still getting the same error.

mafredri:
Obviously prompt is running with incorrect paramteres somewhere in your zsh config. I recommend starting zsh with the debug option and go through the output (it's very verbose) to see where it's happening:

`zsh -x`

jitendravyas:
output of `zsh -x` http://hastebin.com/axizirinet.diff

mafredri:
You are modifying your fpath too late. First zpresto is initialized without and your themes are not yet available in the fpath but you are still setting the theme through presto:

`+/Users/jitendravyas/.zshrc:15> fpath+=( /usr/local/share/zsh/site-functions )`

Then you add the folder to your fpath, reinitialize promptinit and this is why the themes are available.

Not a `pure` bug.

jitendravyas:
There was no fpath in my .zshrc file. I copied from this post https://github.com/sindresorhus/pure/issues/116 and added to my file. 

This was my .zshrc before any modification and it was working fine.

```````````
#
# Executes commands at the start of an interactive session.
#
# Authors:
#   Sorin Ionescu <sorin.ionescu@gmail.com>
#
# Source Prezto.
if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
  source "${ZDOTDIR:-$HOME}/.zprezto/init.zsh"
fi

# Customize to your needs...

``````````````

Now can you help me, what to add in it to make Pure work?
[...]

Issue #120, opened 2015-05-26
Command not found : async


sajadabedi:
Hi.
When i install pure with npm and set config show me this error before run any command  in iterm.
How i can fix this?
![screen shot 1394-03-05 at 11 19 32 pm](https://cloud.githubusercontent.com/assets/5726597/7820437/b8bdab78-03fd-11e5-94ee-cb34861ffda4.png)

mafredri:
To me it looks like `pure` is loading before `async`. Could you go through your `$fpath` and see if you have any other references to `pure` in them?

sajadabedi:
I run command and print this output: @mafredri 
```
/Users/sajad/.oh-my-zsh/plugins/sublime
/Users/sajad/.oh-my-zsh/plugins/git
/Users/sajad/.oh-my-zsh/functions
ls: /Users/sajad/.oh-my-zsh/functions: No such file or directory
/Users/sajad/.oh-my-zsh/completions
ls: /Users/sajad/.oh-my-zsh/completions: No such file or directory
/usr/share/zsh/site-functions
/usr/share/zsh/5.0.5/functions
```

mafredri:
Ok, so looks like you are also using OSX ZSH (not the Homebrew one), so you need to add `/usr/local/share/zsh/site-functions` to your `fpath`

```
fpath+=("/usr/local/share/zsh/site-functions")
```

victormejia:
Having same issue here, pure is running since it was added as a custom theme to oh-my-zsh.

michaelryancaputo:
getting a similar issue:

```/Users/mike/.oh-my-zsh/plugins/cp
/Users/mike/.oh-my-zsh/plugins/bower
/Users/mike/.oh-my-zsh/plugins/npm
/Users/mike/.oh-my-zsh/plugins/brew
/Users/mike/.oh-my-zsh/plugins/git
/Users/mike/.oh-my-zsh/functions
ls: /Users/mike/.oh-my-zsh/functions: No such file or directory
/Users/mike/.oh-my-zsh/completions
ls: /Users/mike/.oh-my-zsh/completions: No such file or directory
/usr/share/zsh/site-functions
/usr/share/zsh/5.0.5/functions
```

alem0lars:
@victormejia @michaelryancaputo @sajadabedi Can you check if both the files `async` and `pure` are in the fpath?

mafredri:
[...]

PR #123, opened 2015-05-27
Zsh integration compatibility fixes

mafredri:
Didn't imagine integrations would break so horribly.

This change should fix problems when integrating with presto an antigen (addresses https://github.com/sindresorhus/pure/issues/118, https://github.com/sindresorhus/pure/issues/120 and https://github.com/sindresorhus/pure/issues/121)

I would appreciate it if everyone with problems (especially issue reporters) could try this branch on their respective zsh-frameworks and see if it fixes the reported problems.

Tested this on prezto already, and seems to work just fine.

### DIFF
--- a/arch/PKGBUILD
+++ b/arch/PKGBUILD
@@ -21,6 +21,4 @@ package() {
   cd $srcdir/$pkgname
   install -Dm644 pure.zsh \
     "$pkgdir/usr/share/zsh/functions/Prompts/prompt_pure_setup"
-  install -Dm644 async.zsh \
-    "$pkgdir/usr/share/zsh/functions/async"
 }
--- a/async.plugin.zsh
+++ b/async.plugin.zsh
@@ -1,1 +1,0 @@
-async.zsh
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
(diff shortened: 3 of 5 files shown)


mafredri:
I have now tested it on both antigen and prezto, both seem to work just fine with this branch. A bit paranoid still that this fix has other repercussions. For example, if someone tries to install this by copying just pure.zsh, that will fail.

An error message that async.zsh is missing, could be printed of course.

sindresorhus:
@mafredri Can you mention the people from the issues? So they can test this.

sindresorhus:
I hate to say this (as I love modularization), but I wonder if it would be easier to just embed `async.zsh` in `pure.zsh`.

mafredri:
@sindresorhus actually I was thinking of rejecting this PR. It was somewhat of a quick fix while everything seemed to be burning. But I've updated the readme with proper instructions for oh-my-zsh and antigen. Still discussing potential solutions to prezto on their issue tracker, but I could update the readme with at least an temporary fix for that as well.

I played with the idea of embedding async in pure as well, but I don't think that's necessary anymore (instructions seem sufficient). Maybe one option would be to release async on npm?

mafredri:
And now prezto has been updated with a fix :)!

sindresorhus:
> And now prezto has been updated with a fix :)!

So I guess this PR can be closed then?

> Maybe one option would be to release async on npm?

I still want async.zsh in this repo to make it easy for people to just clone and use it. But you should definitely publish it on npm regardless.

mafredri:
> I still want async.zsh in this repo to make it easy for people to just clone and use it. But you should definitely publish it on npm regardless.
I agree. And I'll have a look at publishing it on npm.

Closed.

PR #124, opened 2015-05-30
Show hostname in terminal title if session is over ssh

mafredri:
When logged in through ssh to a machine, say `mafredri@Denver`.
* No command active: `Denver: ~`
* Sleep 10: `Denver: mafredri: sleep 10`

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -61,9 +61,13 @@ prompt_pure_check_git_arrows() {
 prompt_pure_preexec() {
 	prompt_pure_cmd_timestamp=$EPOCHSECONDS
 
-	# shows the current dir and executed command in the title when a process is active
+	# tell the terminal we are setting the title
 	print -Pn "\e]0;"
-	echo -nE "$PWD:t: $2"
+	# show hostname if connected through ssh
+	[[ "$SSH_CONNECTION" != '' ]] && print -Pn "%m: "
+	# shows the current dir and executed command in the title when a process is active
+	# (use print -r to disable potential evaluation of escape characters in cmd)
+	print -Pnr "$PWD:t: $2"
 	print -Pn "\a"
 }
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Maybe `Denver: mafredri:` => `mafredri@Denver:` for clarity and it kinda looks better?

mafredri:
@sindresorhus really good suggestion, I like it. But it seems I kind of misunderstood what was already in the title. Since my example was so simple it didn't catch my error in thought :P. If you did `cd /home` the title would actually look like `Denver: home: sleep 10`.

I could also add the username, so `mafredri@Denver: home: sleep 10` (does it get a bit crowded, though?), then we would keep it in line with normal `pure` behavior by showing the path.

sindresorhus:
Yeah, hostname is enough in the title.

mafredri:
I did run it with a while locally as
* `Denver: /home`
* `(Denver) home: sleep 10`

I didn't like the inconsistency, of changing the hostname to parenthesis while a command was running. But maybe it could be `(Denver)` all the time, then we wouldn't get the double `:`.

sindresorhus:
:+1:

sindresorhus:
@mafredri Was thinking of doing another release. Anything else you want to fix / get in beforehand?

What's your npm username? Gonna add you.

mafredri:
Can't think of anything major at this moment. Only thing I'd like to do something about is the excessive grayness when connected over ssh and inside a git repo:
 

But what to do about it I dunno yet :).

Didn't have an npm user beforehand, but created one now [mafredri](https://www.npmjs.com/~mafredri)

sindresorhus:
> Only thing I'd like to do something about is the excessive grayness when connected over ssh and inside a git repo:

Agreed, but not a blocker. Can you open an issue?

> Didn't have an npm user beforehand, but created one now mafredri

Cool, added you. Feel free to do new future releases. Just use the following https://gist.github.com/sindresorhus/8435329 to make sure the release is properly pushed, and do a changelog thingy here: https://github.com/sindresorhus/pure/releases

mafredri:
> Agreed, but not a blocker. Can you open an issue?

Done https://github.com/sindresorhus/pure/issues/126

> Cool, added you. Feel free to do new future releases. Just use the following [...]

Cheers :)! Sure, will do.

PR #130, opened 2015-06-13
Rename two internal variables for readability

zmwangx:
This PR is highly subjective, but anyway, I found two confusingly named (to me) variables that caused me a bit of trouble when reading the code — I had to go back and forth between several functions to understand what they are doing. Here they are:

* `prompt_pure_git_delay_dirty_check` sounds like a boolean, but holds the (finish) timestamp of the last dirty check (if the dirty check took more than 2 seconds).

  I renamed it to `prompt_pure_git_last_dirty_check_timestamp`.

* Local variable `dirty_check` holds the elapsed time in seconds since the last dirty check (if the timestamp is set), but the variable name is too generic to make sense of.

  I renamed it to `time_since_last_dirty_check`.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -83,7 +83,7 @@ prompt_pure_preprompt_render() {
 
 	# set color for git branch/dirty status, change color if dirty checking has been delayed
 	local git_color=242
-	[[ -n ${prompt_pure_git_delay_dirty_check+x} ]] && git_color=red
+	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# construct prompt, beginning with path
 	local prompt="%F{blue}%~%f"
@@ -192,7 +192,7 @@ prompt_pure_async_tasks() {
 
 		# reset git preprompt variables, switching working tree
 		unset prompt_pure_git_dirty
-		unset prompt_pure_git_delay_dirty_check
+		unset prompt_pure_git_last_dirty_check_timestamp
 
 		# set the new working tree, prefixed with "x"
 		prompt_pure_current_working_tree="x${working_tree}"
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -227,9 +227,9 @@ prompt_pure_async_callback() {
 			prompt_pure_git_dirty=$output
 			prompt_pure_preprompt_render
 
-			# when prompt_pure_git_delay_dirty_check is set, the git info is displayed in a different color, this is why the
-			# prompt is rendered before the variable is (potentially) set
-			(( $exec_time > 2 )) && prompt_pure_git_delay_dirty_check=$EPOCHSECONDS
+			# when prompt_pure_git_last_dirty_check_timestamp is set, the git info is displayed in a different color,
+			# this is why the prompt is rendered before the variable is (potentially) set
```
mafredri: Hmm, your editing this comment made me realize that it itself is very confusing. I think it would be best to remove the whole `this is why the prompt is rendered before the variable is (potentially) set` since prompt rendering is not a result of this variable being set or not.

mafredri:
Good catch with the variable names, I agree that the renamed ones make a lot more sense, thanks! Would you mind still modifying the one comment I commented on?

pure.zsh
zmwangx: @mafredri Hmm, but I think it is? `git_color` will be `red` if this variable is set (see https://github.com/zmwangx/pure/blob/5b522d86dad019b08e2684114fc9716266d2ffc6/pure.zsh#L85-L86).

mafredri: No, not in the sense that the comment is saying. I think something along these lines is more appropriate:
```sh
# When prompt_pure_git_delay_dirty_check_timestamp is set, the git info is displayed in a different color.
# To distinguish between a "fresh" and a "cached" result, the preprompt is rendered before setting this
# variable. Thus, only upon next rendering of the preprompt will the result appear in a different color.
```

zmwangx:
I thought the original comment was already pretty understandable (I read the comment, went back to render, and immediately understood what was going on), but sure, why not. Changed and rebased. Any other problems? I see that you were using a slightly different variable name from mine in the previous comment (`prompt_pure_git_delay_dirty_check_timestamp` as opposed to `prompt_pure_git_last_dirty_check_timestamp`).

mafredri:
Hmm, you are right. Now that I look at the old comment again it makes sense, momentary lapse in thought, perhaps? :). The new comment is perhaps overly verbose, but I hope it's more clear.

Regarding `prompt_pure_git_delay_dirty_check_timestamp` vs `prompt_pure_git_last_dirty_check_timestamp` I should have just copy-pasted your variable name instead of adding `_timestamp` to the old one.

Thanks for the PR!

zmwangx:
> Now that I look at the old comment again it makes sense, momentary lapse in thought, perhaps?

Yeah, happens all the time ;)

PR #138, opened 2015-07-07
Simplifies loading Pure in zgen by adding symbolic link to async

blai:
I use [zgen](https://github.com/tarjoilija/zgen) to load Pure.
Simply doing `zgen load sindresorhus/pure` would not work, and will see the following errors:

```
prompt_pure_setup:10: async: function definition file not found
prompt_pure_async_tasks:3: command not found: async_start_worker
prompt_pure_async_tasks:4: command not found: async_register_callback
```

Adding this symbolic link make it work (without having to do those manual steps in the README).

### DIFF
--- a/async.plugin.zsh
+++ b/async.plugin.zsh
@@ -1,0 +1,1 @@
+async.zsh


forivall:
This will break antigen, since antigen will only load the first `.plugin.zsh` file found. Just add `zgen load mafredri/zsh-async` before loading pure.

blai:
@forivall the suggested solution works well, thanks. :+1:

PR #142, opened 2015-07-13
prompt_pure_string_length: no need to subtract one

zmwangx:
`prompt_pure_string_length` is used for calculating the length of the preprompt, but the preprompt doesn't contain a newline. This seems to be leading to off-by-one errors in prompt length calculations (which doesn't seem to have dire consequences, though).

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -111,7 +111,7 @@ prompt_pure_preprompt_render() {
 
 		# disable clearing of line if last char of preprompt is last column of terminal
 		local clr="\e[K"
-		(( $preprompt_length * $lines == $COLUMNS - 1 )) && clr=""
+		(( $COLUMNS * $lines == $preprompt_length )) && clr=""
 
 		# modify previous preprompt
 		print -Pn "\e7\e[${lines}A\e[1G${preprompt}${clr}\e8"
--- a/pure.zsh
+++ b/pure.zsh
@@ -76,8 +76,7 @@ prompt_pure_preexec() {
 
 # string length ignoring ansi escapes
 prompt_pure_string_length() {
-	# Subtract one since newline is counted as two characters
-	echo $(( ${#${(S%%)1//(\%([KF1]|)\{*\}|\%[Bbkf])}} - 1 ))
+	echo $(( ${#${(S%%)1//(\%([KF1]|)\{*\}|\%[Bbkf])}} ))
 }
 
 prompt_pure_preprompt_render() {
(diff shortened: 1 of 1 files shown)


mafredri:
When you remove the `-1` this happens:


When really it should look like this:

zmwangx:
You're right. But now that I read the code more closely, I realized that the reason minus one works is not `# Subtract one since newline is counted as two characters` as seen in the original comment (maybe I misinterpreted that); `prompt_pure_string_length` isn't the right place for the minus one. Instead, the number of lines the preprompt occupies should be

```zsh
$(( ($prompt_length - 1) / $COLUMNS + 1 ))
```

so the right place to subtract one is when we calculate `lines`. I've updated the commit accordingly.

zmwangx:
By the way, I also just noticed another error: [L122](https://github.com/sindresorhus/pure/blob/579af2425b58483f1343afb97bf9ea12efb85d3b/pure.zsh#L112)

```zsh
(( $prompt_length * $lines == $COLUMNS - 1 )) && clr=""
```
The condition is supposed to capture the case where the last char of prompt is last column of terminal, so it should instead be

```zsh
(( $COLUMNS * $lines == $prompt_length )) && clr=""
```

(that's after moving the minus one, which is the logical thing to do; otherwise we need to write `$prompt_length + 1` here).

mafredri:
> so the right place to subtract one is when we calculate lines. I've updated the commit accordingly.

You are absolutely right, that comment seems to have been a left-over from earlier versions when the newline was included. Afterwards I've just compensated for it without really thinking twice about it.

> The condition is supposed to capture the case where the last char of prompt is last column of terminal, so it should instead be

Yes! Good catch, your proposed solutions seems correct when combined with the changes already present in this PR.

zmwangx:
> that comment seems to have been a left-over from earlier versions when the newline was included.

That makes more sense now. I should have got-blamed it.

zmwangx:
Resolved merge conflict and rebased. Hopefully doing the right thing.

I also took the liberty to add a commit to address the other bug mentioned in https://github.com/sindresorhus/pure/pull/142#issuecomment-120961475.

mafredri:
Thanks for this! I've confirmed it works as expected.

This logic could still potentially still be improved by calculating if any lines above need to be erased based on the previous preprompt length and current columns in the terminal. The output would be prettier, now there can potentially be left a "dead line" (or more) of output above the new preprompt.

zmwangx:
> This logic could still potentially still be improved by calculating if any lines above need to be erased based on the previous preprompt length and current columns in the terminal.

At the expense of potentially shifting the current line, though.

mafredri:
> At the expense of potentially shifting the current line, though.

How do you mean?

zmwangx:
Ah sorry, misinterpretation, never mind. At a glance, I thought you were saying we should calculate the shifts using the previous preprompt length.

> calculating if any lines above need to be erased 

That's a good idea.

mafredri:
No worries, I created an issue for it so I don't forget :) https://github.com/sindresorhus/pure/issues/145

PR #144, opened 2015-07-13
prompt_pure_preprompt_render: rename prompt to preprompt

zmwangx:
This PR is highly subjective.

Preprompt and prompt are two distinct concepts in pure (or so I believe): the preprompt is where the directory and git info live, whereas the prompt (`PROMPT`) is as pure and simple as `(?.%F{magenta}.%F{red})❯%f`.

In the `prompt_pure_preprompt_render` function though, prompt and preprompt are used interchangeably, which makes it confusing the first time one reads it. (The lines `[[ "${prompt_pure_last_preprompt}" != "${prompt}" ]]` and `prompt_pure_last_preprompt=$prompt` perfectly illustrate the potential point of confusion.)

Therefore, this commits corrects the problem by renaming the `prompt` local variable to `preprompt`, and updating comments accordingly.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -78,45 +78,45 @@ prompt_pure_string_length() {
 }
 
 prompt_pure_preprompt_render() {
-	# check that no command is currently running, the prompt will otherwise be rendered in the wrong place
+	# check that no command is currently running, the preprompt will otherwise be rendered in the wrong place
 	[[ -n ${prompt_pure_cmd_timestamp+x} && "$1" != "precmd" ]] && return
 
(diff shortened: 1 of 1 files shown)


mafredri:
Good points! Even details like these are important :+1:. I don't mind merging, but I believe it might break your other PR.

zmwangx:
I can resolve the merge conflicts if and when the other PR gets merged.

mafredri:
Great! :)

zmwangx:
Thanks. I'll resolve and rebase the other one after dinner.

Issue #145, opened 2015-07-15
Reduce visual glitches when pre-prompt is same length as the terminal has columns


mafredri:
When the preprompt line of pure is dangerously close to `$COLUMNS` of the terminal, the following can happen when updating it.

![](https://www.dropbox.com/s/n1gco44l5n3fxgs/Screenshot%202015-07-15%2016.22.15.png?dl=1)

There are two edge cases that are triggered when async jobs are done, and the pre-prompt is rendered (as can be seen in the screenshot):

1. Pre-prompt length goes from `$length <= $COLUMNS` to `$length > $COLUMNS`. This results in the spacing between commands disappearing.
2. Pre-prompt length goes from `$length > $COLUMNS` to `$length <= $COLUMNS`. This results in a "dead line" above the pre-prompt, with part of the old pre-prompt remaining.

**Potential solution**

Store the length of the previous pre-prompt, compare it to current `$COLUMNS` and for `1.` decide if we need to add one or more newlines at top, and for `2.` decide if you should erase any lines above the new pre-prompt.

zmwangx:
While we are at it, I would mention a related area where the place to update/whether to update can be fine-grained, namely, when the user clears screen before the preprompt is updated. See screencast:

[![asciicast](https://asciinema.org/a/23484.png)](https://asciinema.org/a/23484)

To address this, we need to detect the absolute terminal line we are on.

Do you want a separate issue for this?

mafredri:
I don't mind keeping it to one issue, even if this is a slightly different problem. Something like this could be used to know if the screen has been cleared. But I'm not so sure I like hijacking the users keybindings :/. Can we get any other useful details out of ZLE re: which line the cursor is on?

```zsh
typeset -gi clear
capture-clear() {
	zle clear-screen
	clear=1
}
zle -N capture-clear
bindkey '^l' capture-clear
```

mafredri:
Actually, this clearing case seems pretty simple unless you can find any negatives / downsides with this implementation (works without any extra modifications)

```zsh
prompt_pure_clear_screen() {
	zle .clear-screen
	# enable output to terminal
	zle -I
	# clear the initial prompt and move cursor to beginning
	print -Pn "\e[1A\e[1G\e[K"
	# print precmd
	prompt_pure_preprompt_render precmd
}
zle -N clear-screen prompt_pure_clear_screen
```

zmwangx:
> I don't mind keeping it to one issue, even if this is a slightly different problem.

[...]

PR #147, opened 2015-07-18
Preserve preprompt on Ctrl+L

mafredri:
As discussed in #145 This PR fixes clear screen in pure (where the preprompt is missing when issuing Ctrl+L).

Due to third-party libraries doing whatever they want, I've decided to take a more fragile, but safer approach in this PR (see https://github.com/sindresorhus/pure/commit/4961520addf1afe298b07cf1f4e32c85fa083ca3 for details).

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -47,6 +47,15 @@ prompt_pure_check_cmd_exec_time() {
 	(($elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5})) && prompt_pure_human_time $elapsed
 }
 
+prompt_pure_clear_screen() {
+	# enable output to terminal
+	zle -I
+	# clear screen and move cursor to (0, 0)
+	print -n "\e[2J\e[0;0H"
+	# print preprompt
+	prompt_pure_preprompt_render precmd
+}
+
 prompt_pure_check_git_arrows() {
 	# check if there is an upstream configured for this branch
 	command git rev-parse --abbrev-ref @'{u}' &>/dev/null || return
@@ -249,6 +258,7 @@ prompt_pure_setup() {
 	prompt_opts=(subst percent)
 
 	zmodload zsh/datetime
+	zmodload zsh/zle
 	autoload -Uz add-zsh-hook
 	autoload -Uz vcs_info
 	autoload -Uz async && async
@@ -260,6 +270,13 @@ prompt_pure_setup() {
 	zstyle ':vcs_info:*' use-simple true
 	zstyle ':vcs_info:git*' formats ' %b'
 	zstyle ':vcs_info:git*' actionformats ' %b|%a'
+
+	# if the user has not registered a custom zle widget for clear-screen,
+	# override the builtin one so that the preprompt is displayed correctly when
+	# ^L is issued.
+	if [[ $widgets[clear-screen] == "builtin" ]]; then
+		zle -N clear-screen prompt_pure_clear_screen
+	fi
 
 	# show username@host if logged in through SSH
 	[[ "$SSH_CONNECTION" != '' ]] && prompt_pure_username=' %F{242}%n@%m%f'
--- a/pure.zsh
+++ b/pure.zsh
@@ -51,7 +51,7 @@ prompt_pure_clear_screen() {
 	# enable output to terminal
 	zle -I
 	# clear screen and move cursor to (0, 0)
-	print -n "\e[2J\e[0;0H"
+	print -n '\e[2J\e[0;0H'
 	# print preprompt
 	prompt_pure_preprompt_render precmd
 }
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -153,6 +153,19 @@ antigen bundle sindresorhus/pure
 
 ## FAQ
 
+### My preprompt is missing when I clear the screen with Ctrl+L
+
+Pure doesn't register it's custom *clear-screen* widget if it has been previously modified. If you haven't registered your
```
sindresorhus: it's => its

readme.md
```diff
@@ -153,6 +153,19 @@ antigen bundle sindresorhus/pure
...
+own zle widget with `zle -N clear-screen custom-clear-screen` it might have been done by third-party modules. For
+example `zsh-syntax-highlighting` and `zsh-history-substring-search` are known to do this and they should for that
+reason be **the very last thing** in your `.zshrc` (as pointed out in their documentation).
+
+To find out the culprit that is overriding your *clear-screen* widget, you can run the following command:
+
+```shell
+zle -l | grep clear-screen
+```
+
```
sindresorhus: Don't use hard wrapping.

mafredri: Alright, I'll do it inline

readme.md
mafredri: Thanks! :)

readme.md
mafredri: I assumed at least you meant the code-block, or do you mean the text also?

sindresorhus: I meant everything. No reason to do hard-wrapping anymore. All editors supports soft-wrapping, even GitHub diffs does.

mafredri: I don't do hard-wrapping in my editor, but I don't mind changing it (although the rest of the readme is hard-wrapped).

mafredri:
Amended comments on readme and did a force push of the branch.

pure.zsh
```diff
@@ -47,6 +47,15 @@ prompt_pure_check_cmd_exec_time() {
 	(($elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5})) && prompt_pure_human_time $elapsed
 }
 
+prompt_pure_clear_screen() {
+	# enable output to terminal
+	zle -I
+	# clear screen and move cursor to (0, 0)
+	print -n "\e[2J\e[0;0H"
```
sindresorhus: Could this use single-quotes? I kinda like to use single-quotes when there's no variable expansion going on to make the intent clear. Does that make sense? (I'm still not sure about it).

pure.zsh
```diff
@@ -261,6 +271,13 @@ prompt_pure_setup() {
 	zstyle ':vcs_info:git*' formats ' %b'
 	zstyle ':vcs_info:git*' actionformats ' %b|%a'
 
+	# if the user has not registered a custom zle widget for clear-screen,
+	# override the builtin one so that the preprompt is displayed correctly when
+	# ^L is issued.
+	if [[ $widgets[clear-screen] == "builtin" ]]; then
```
sindresorhus: same as https://github.com/sindresorhus/pure/pull/147/files#r34948206

readme.md
```diff
@@ -153,6 +153,12 @@ antigen bundle sindresorhus/pure
 
 ## FAQ
 
+### My preprompt is missing when I clear the screen with Ctrl+L
+
+Pure doesn't register its custom *clear-screen* widget if it has been previously modified. If you haven't registered your own zle widget with `zle -N clear-screen custom-clear-screen` it might have been done by third-party modules. For example `zsh-syntax-highlighting` and `zsh-history-substring-search` are known to do this and they should for that reason be **the very last thing** in your `.zshrc` (as pointed out in their documentation).
+
+To find out the culprit that is overriding your *clear-screen* widget, you can run the following command: `zle -l | grep clear-screen`.
+
```
sindresorhus: This section is awesome. Nice work.

Though it's funny how creating a ZSH prompts ends up having to document most of ZSH :p

sindresorhus:
Tried it now. Works good. I always bothered me that the preprompt would disappear and I had no idea it was fixable :)

pure.zsh
mafredri: Single-quotes would work, no doubt. Although, when working with shell scripts I kind of prefer double quotes. It's easier to add a variable later (no need to convert quotes) and you don't really need to think about which to use. The exception I usually do is when you have double-quotes inside the string, then I might wrap it in single-quotes.

I'm fine with either as long as we stick to one way of doing it :)!

pure.zsh
mafredri: Actually, this doesn't even need quotes, but I usually do it anyway just to indicate what it is :/.

readme.md
mafredri: > This section is awesome. Nice work.

Thanks!

> Though it's funny how creating a ZSH prompts ends up having to document most of ZSH :p

Haha, that's so true it almost hurts! :D

pure.zsh
sindresorhus: Yeah. I agree. Better to be explicit.

pure.zsh
sindresorhus: Ok. Let's stick with single-quotes wherever possible for now then. We can always change later if that feels weird.

mafredri: Alright, I'll update the PR.

mafredri:
Yeah, I was kind of bothered by it too, and if @zmwangx hadn't brought it up, I probably wouldn't have investigated it further, so thanks to him (and his input)! :)

PR #148, opened 2015-07-19
Underline repo name in path display

forivall:
I'm not sure if this is something that should be put into the repository, it's just an enhancement that I personally like, and have had sitting around for a while (ignore the commit timestamp). This also has a bunch of special casing that definitely would need to be cleaned up before it could be merged, if it's something we should merge at all. And there's probably a better way to perform the string manipulation; I wanted to make sure I didn't fork out to keep it fast, and used the only string manipulation hammer I know (`#` and `%`).

zsh-pure-underline

obligatory really-big-cuz-that-makes-it-pretty screenshot

zsh-pure-underline-big-2

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -98,7 +98,7 @@ prompt_pure_print_path() {
        if [[ $repo != $upath || $vcs_info_msg_2_ == '.' ]]; then
            local parent="${repo:h}/"
            # if parent is ./ we can safely remove it
-           local _repo="${parent#./}%U${repo:t}%u"
+           local _repo="${parent#./}%B${repo:t}%b%F{blue}"
            upath=${upath/$repo/$_repo}
        fi
    fi
--- a/pure.zsh
+++ b/pure.zsh
@@ -47,6 +47,55 @@ prompt_pure_check_cmd_exec_time() {
 	(($elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5})) && prompt_pure_human_time $elapsed
 }
 
+if [[ "$PURE_HIGHLIGHT_REPO" == 1 ]] ; then
+	# underline
+	PURE_UNDERLINE_START="$(tput smul)"
+	PURE_UNDERLINE_END="$(tput rmul)"
+	# bold
+	# PURE_UNDERLINE_START=${fg_bold[blue]}
+	# PURE_UNDERLINE_END=${fg_no_bold[blue]}
+	prompt_pure_render_preprompt_path() {
+		if [[ -n "$vcs_info_msg_1_" ]] ; then
+			local prerepo reponame postrepo
+			if [[ "$PWD" == "$vcs_info_msg_1_"* ]]; then
+				# the working directory is the real path
+				prerepo=${vcs_info_msg_1_%/*}
+				if [[ "$prerepo" == "$HOME"* ]]; then prerepo="~${prerepo#$HOME}"; fi # replace ~
+				prerepo="${prerepo%/}/"
+				reponame=${vcs_info_msg_1_##*/}
+				postrepo="${PWD#$vcs_info_msg_1_}"
+				postrepo=${postrepo#/}
+				[[ -n "${postrepo}" ]] && postrepo="/${postrepo}"
+				prompt_pure_preprompt_path="%F{blue}${prerepo}${PURE_UNDERLINE_START}${reponame}${PURE_UNDERLINE_END}${postrepo}"
+			elif [[ "$PWD" == */"$vcs_info_msg_2_" ]]; then
+				# the working directory is in a git repository under a symlink
+				local repopath="${PWD%/$vcs_info_msg_2_}"
+				reponame=${repopath##*/}
+				prerepo=${repopath%/*}
+				if [[ "${prerepo}" == "$HOME"* ]]; then prerepo="~${prerepo#$HOME}" fi # replace ~
+				prerepo="${prerepo%/}/"
+				prompt_pure_preprompt_path="%F{blue}${prerepo}${PURE_UNDERLINE_START}${reponame}${PURE_UNDERLINE_END}/$vcs_info_msg_2_"
+			elif [[ "$vcs_info_msg_2_" == "." ]]; then
+				# the working directory is at the root of a repo under a symlink
+				prerepo=${PWD%/*}
+				if [[ "${prerepo}" == "$HOME"* ]]; then prerepo="~${prerepo#$HOME}" fi # replace ~
+				prerepo="${prerepo%/}/"
+				reponame=${PWD##*/}
+				prompt_pure_preprompt_path="%F{blue}${prerepo}${PURE_UNDERLINE_START}${reponame}${PURE_UNDERLINE_END}"
+			else
+				# fallback: the working directory symlinks into a repo somewhere else
+				prompt_pure_preprompt_path="%F{blue}%~"
+			fi
+		else
+			prompt_pure_preprompt_path="%F{blue}%~"
+		fi
+	}
+else
+	prompt_pure_render_preprompt_path() {
+		prompt_pure_preprompt_path="%F{blue}%~"
+	}
+fi
+
 prompt_pure_clear_screen() {
 	# enable output to terminal
 	zle -I
@@ -97,7 +146,9 @@ prompt_pure_preprompt_render() {
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# construct preprompt, beginning with path
-	local preprompt="%F{blue}%~%f"
+	local prompt_pure_preprompt_path
+	prompt_pure_render_preprompt_path
+	local preprompt="$prompt_pure_preprompt_path"
 	# git info
 	preprompt+="%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f"
 	# git pull/push arrows
@@ -266,10 +317,11 @@ prompt_pure_setup() {
 	add-zsh-hook precmd prompt_pure_precmd
 	add-zsh-hook preexec prompt_pure_preexec
 
+	zstyle ':vcs_info:*' max-exports 3
 	zstyle ':vcs_info:*' enable git
 	zstyle ':vcs_info:*' use-simple true
-	zstyle ':vcs_info:git*' formats ' %b'
-	zstyle ':vcs_info:git*' actionformats ' %b|%a'
+	zstyle ':vcs_info:git*' formats ' %b' '%R' '%S'
+	zstyle ':vcs_info:git*' actionformats ' %b|%a' '%R' '%S'
 
 	# if the user has not registered a custom zle widget for clear-screen,
 	# override the builtin one so that the preprompt is displayed correctly when
--- a/pure.zsh
+++ b/pure.zsh
@@ -47,55 +47,6 @@ prompt_pure_check_cmd_exec_time() {
 	(($elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5})) && prompt_pure_human_time $elapsed
 }
 
-if [[ "$PURE_HIGHLIGHT_REPO" == 1 ]] ; then
-	# underline
-	PURE_UNDERLINE_START="$(tput smul)"
-	PURE_UNDERLINE_END="$(tput rmul)"
-	# bold
-	# PURE_UNDERLINE_START=${fg_bold[blue]}
-	# PURE_UNDERLINE_END=${fg_no_bold[blue]}
-	prompt_pure_render_preprompt_path() {
-		if [[ -n "$vcs_info_msg_1_" ]] ; then
-			local prerepo reponame postrepo
-			if [[ "$PWD" == "$vcs_info_msg_1_"* ]]; then
-				# the working directory is the real path
-				prerepo=${vcs_info_msg_1_%/*}
-				if [[ "$prerepo" == "$HOME"* ]]; then prerepo="~${prerepo#$HOME}"; fi # replace ~
-				prerepo="${prerepo%/}/"
-				reponame=${vcs_info_msg_1_##*/}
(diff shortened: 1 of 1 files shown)


mafredri:
I don't completely hate this idea, my initial impression was, **nooo**. But I will give it a chance. I do have some points to raise about this PR though.

* It seems overly complicated for a simple feature (see below)
* The way you are evaluating the option means that it cannot be configured _after_ `pure` has been initialized
* Why not use the zsh own syntax for underlining `%U` and `%u`
* What do the changes to vcs_info do?

**Note** this patch is VERY naive and doesn't take into account paths such as `~/git/git` where the last `git` is the repo, it would underline the first git (can be worked around, but still).

```diff
diff --git a/pure.zsh b/pure.zsh
index 45840aa..8c74617 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -96,8 +96,14 @@ prompt_pure_preprompt_render() {
        local git_color=242
        [[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
+       local upath='%~'
        # construct preprompt, beginning with path
-       local preprompt="%F{blue}%~%f"
+       if [[ -n $prompt_pure_current_working_tree ]]; then
+               local reponame=$(basename $prompt_pure_current_working_tree)
+               upath=${${(%)upath}/$reponame/'%U'$reponame'%u'}
+       fi
+
+       local preprompt="%F{blue}${upath}%f"
        # git info
        preprompt+="%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f"
        # git pull/push arrows
```

mafredri:
Ok, so I figured out what the `vcs_info:git` changes are for, basically `%R` is full path to repo, `%S` is relative path inside repo and `max-exports` provides us with n-amount of `$vcs_info_msg_n` variables. Nifty. We could probably use this elsewhere in pure instead of `command git rev-parse --show-toplevel`.

forivall:
Yeah, opening subshells isn't really a problem, it's using sed and awk and such that often causes slowdowns. And I do admit, in its current form, it is overly complicated and very naive. I'll be looking to see if there's a better way to do this; it definitely shouldn't be merged in its current form.

Really, I just want to know if this is a feature that could be included in the core, or if i should just keep this as my personal fork and figure out how to make it work in a more external, independent fashion.

mafredri:
After using this for a day I kind of like it. But I'm a bit torn, at the same time I think it perhaps sticks out a bit much. It's definitely different, and I think there can be pretty strong opinions on it.

Btw, I hope you don't mind, but I got a bit inspired and went ahead and rewrote this a bit:

```patch
diff --git a/pure.zsh b/pure.zsh
index 45840aa..94ff65a 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -88,6 +88,24 @@ prompt_pure_string_length() {
 	echo $(( ${#${(S%%)1//(\%([KF1]|)\{*\}|\%[Bbkf])}} ))
 }
 
+prompt_pure_print_path() {
+	local upath='%~'
+	if [[ -n $vcs_info_msg_1_ ]]; then
+		# expand %~
+		upath=${(%)upath}
+		# remove git subdirectory from path to get "pretty" repo path
+		local repo=${upath%'/'$vcs_info_msg_2_}
+		if [[ $repo != $upath || $vcs_info_msg_2_ == '.' ]]; then
+			local parent="${repo:h}/"
+			# if parent is ./ we can safely remove it
+			local _repo="${parent#./}%U${repo:t}%u"
+			upath=${upath/$repo/$_repo}
+		fi
+	fi
+
+	print $upath
+}
+
 prompt_pure_preprompt_render() {
 	# check that no command is currently running, the preprompt will otherwise be rendered in the wrong place
 	[[ -n ${prompt_pure_cmd_timestamp+x} && "$1" != "precmd" ]] && return
@@ -97,7 +115,7 @@ prompt_pure_preprompt_render() {
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# construct preprompt, beginning with path
-	local preprompt="%F{blue}%~%f"
+	local preprompt="%F{blue}$(prompt_pure_print_path)%f"
 	# git info
 	preprompt+="%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f"
 	# git pull/push arrows
@@ -266,10 +284,11 @@ prompt_pure_setup() {
 	add-zsh-hook precmd prompt_pure_precmd
 	add-zsh-hook preexec prompt_pure_preexec
 
+	zstyle ':vcs_info:*' max-exports 3
-	zstyle ':vcs_info:git*' formats ' %b'
-	zstyle ':vcs_info:git*' actionformats ' %b|%a'
+	zstyle ':vcs_info:git*' formats ' %b' 'x%R' '%S'
+	zstyle ':vcs_info:git*' actionformats ' %b|%a' 'x%R' '%S'
 
 	# if the user has not registered a custom zle widget for clear-screen,
 	# override the builtin one so that the preprompt is displayed correctly when
```

In action:

forivall:
@mafredri I don't mind at all; your re-write is great, and way cleaner than my original. The only thing I would do differently is that I prefer to pass values by setting an out variable instead of printing it and capturing with `$()`, but that's mainly so that it's easier to print debug messages. And I'm superstitious about subshells. But that's just superstition :stuck_out_tongue:.

I liked it so much that I committed it for this PR.

mafredri:
> The only thing I would do differently is that I prefer to pass values by setting an out variable instead of printing it and capturing with `$()`, but that's mainly so that it's easier to print debug messages.

It's not bad in my opinion, I just prefer less variables (and there's the downside of having to call a function before the variable is available.

> And I'm superstitious about subshells. But that's just superstition :stuck_out_tongue:.

Well... I decided to check it out, and you should be! Huge performance loss with subshell vs variable: https://gist.github.com/mafredri/25aa77de809e86d24a63

> I liked it so much that I committed it for this PR.

Awesome!

sindresorhus:
Any thoughts on using `bold` instead of `underline`?

forivall:
I originally used bold, but it looked bad on cygwin, which I was using when I did my naive implementation. Otherwise, I'm just as happy with bold, which looks fine in my usual linux terminal. 
zsh-pure-underline-big-3

sindresorhus:
How does it look bad on Cygwin? Maybe open an issue about that there so it can eventually be fixed.

forivall:
It's fine, it was just mediocre kerning, nothing that should really be an issue for anyone but me.

mafredri:
Bold doesn't look good for me either, and imo it doesn't quite fit along with the normal width text, especially when you're in a subdir of the repo, it just looks weird :/. I don't even usually have bold fonts enabled in Terminal because they never looked good.

sindresorhus:
I'm fine with underline too btw. Would just have preferred bold.

mafredri:
With bolds you can also select "Use bright colors fold bold test" in Terminal, so it's possible to do stuff like this (by modifying your bright colors):



Thought it was worth a mention, as bold is definitely more versatile.

sindresorhus:
@mafredri Go with whatever you think is best. You're the maintainer. I trust your judgement ;)

mafredri:
Hehe, thanks for the vote of confidence @sindresorhus ;).

I'll try out this bold business for a day and make a decision after that :+1: !

mafredri:
I still have a bit of bias towards the underline, but I think bold is the more sensible choice here. Even though I love the underlined git repo, I can't shake the feeling that it's too "in yo face".

@forivall is there anything you'd like to address before this gets merged? I will likely make some small modifications to this when rebasing my `pure-nitro` branch thought (so re: subshells you don't need to worry about that).

forivall:
Nope, I'm fine with this as it is, and I'm just as happy with bold or underline. The whole point is that the root of the git repo is easy to see, and I don't have any strong opinions on how it happens. Other than blinking. No blinking.

zmwangx:
@mafredri Would you consider making this feature optional? I know having a lot of customization parameters goes against the general principles of `pure`, but this one is a pretty significant change and definitely have broader consequences than, say, `PURE_GIT_DOWN_ARROW`.

I don't want to downplay this idea since I know some people will like it, but I'm not a fan because it

1. adds very little useful information — 99.5% of the time I know exactly where the git root is just by glancing at PWD (it's *my* filesystem after all);
2. stands out too much (all for nothing, most of the time);
3. makes `pure` less "pure" — currently everything serves a purpose, but this one seems redundant to a large extent.

While I'm not a `pure` user (I use a similar "backend"), I could certainly foresee many actual `pure` users share my discontent. Your feelings totally check out with me:

> I don't completely hate this idea, my initial impression was, **nooo**.

(I saw this a few days ago, and to this day my impression is still nooo.)

> It's definitely different, and I think there can be pretty strong opinions about it.

> I can't shake the feeling that it's too "in yo face".

zmwangx:
By the way, using bold is probably not a good idea, since bold depends on terminal handling. Bold may be translated to bright on the terminal level, and depending on color theme users might see strange results; for instance, with the Solarized theme, bright counterparts are almost not related to originals at all, and they are generally dull colors (I'm talking about Solarized's `base*` colors, which are usually mapped to `8-15` in 16-color terminal color themes). Users shouldn't have to modify terminal behavior (interpretation of `\e[1m`) and predefined color themes to get a reasonable prompt.

My conclusion: if we were to use bold, then that alone makes an even stronger case for the ability to turn this feature off.

mafredri:
@forivall alright, goodie!

@zmwangx well put, and you do raise some valid points. I don't exactly agree that it's redundant though, I can quickly glance how deep I'm inside a git working tree without even a need to consider which project it is. And while I don't see this providing much useful information either, I feel it's more of a "theme" thing, themes make things look fancy.

But I also do share some of your concerns, esp. regarding handling of bold (even though it's quite controllable from terminal settings usually, users likely don't know that).

If I were to make this optional, it would be through e.g. `PURE_GIT_HIGHLIGHT_MODE` where `0 = disabled`, `1 = underline` and `2 = bold` for example.

Once again I'm torn, I'll try to sleep on this and see if I have some clarity.

zmwangx:
> even though it's quite controllable from terminal settings usually, users likely don't know that

Even if users know that, chances are they shouldn't/don't want to modify the settings:

* Bold: some folks may disable bold and instead translate to bright for a reason (and certain lame terminals might not even support bold);
* Color theme: while individual colors customizable, it's a bad idea to change one color in a pack, and random changes break other things, e.g., editor color themes.

sindresorhus:
I've been thinking about this for a while and I don't think I would use this. I can't imagine any time I would need a visual marker for where the git root is. Also as @zmwangx mentioned, git can be run in any subfolder. Kinda feels like this would be a better fit as some kind of ZSH module users could use with any theme.

-

Sidenote: It's super annoying that you can't assume `bold` will actually be `bold`... This is why I hate shells. Everything is so unreliable, unconsistent and fragile.

zmwangx:
> Everything is so unreliable, unconsistent and fragile.

When you have more than forty years of legacy, with every vendor doing whatever they damn well please...

forivall:
guh, all of this makes me wish that there was a `zstyle` to control [Prompt Expansion](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html)

sindresorhus:
> guh, all of this makes me wish that there was a zstyle to control Prompt Expansion

Sounds like a good issue to open on the ZSH issue tracker ;)

zmwangx:
> > guh, all of this makes me wish that there was a zstyle to control Prompt Expansion
>
> Sounds like a good issue to open on the ZSH issue tracker ;)

I have to disagree.

First, prompt expansion is a core feature while `zstyle` is from the `zsh/zutil` module. There's no way a core feature should be modifiable by an optional module...

More importantly, what @forivall wants to achieve is clearly (quite easily) achievable with existing tools, as demonstrated in this PR, so there's no justification for adding new features of questionable value to the core. The builtin prompt escapes are there to make users' lives easier, but they are by no means necessary since all other expansions are available, including arbitrary function and external command invocations; **having more customizability to the most basic escapes will only make lives of extension authors harder** (you know: arbitrary userland customizations — often done in configuration frameworks without the average user's knowledge and explicit consent — break all assumptions, and make writing foolproof code and debugging as hard as hell).

Zsh is already considered bloated by many (disclaimer: not me), so let's just try to do amazing things with what we already have.

sindresorhus:
@zmwangx I stand corrected. I strongly agree with everything you just said.

forivall:
I agree too; my thought process was just like, "zstyle can control, like anything" "zstyle should be used to control print!" "???" "profit!". Yeah, it's a built in.

I'm 100% ok with rejection of this PR, I'll just maintain it as a continually updated fork, and, if it fancies me at some point, make a `zprint` function that provides an extensible alternative to `print -P`. (and it'll provide a monkey :monkey_face: patch for `print` :scream: :see_no_evil:)

zmwangx:
> and it'll provide a monkey :monkey_face: patch for print :scream: :see_no_evil:

Well, you can always override the builtin `print`, since few people are paranoid enough to write `builtin print` :stuck_out_tongue:

mafredri:
I had kind of hoped this could've made a nice addition to pure, but in light of all the crossed opinions on this (including my own), let's just keep it out. Sorry @forivall :).

-
@forivall here's that monkey-patch you were requesting ;). You'd still have to modify `vcs_repo` behavior after pure init though:

```zsh
print() {
	if [[ ${funcstack[2]} == 'prompt_pure_preprompt_render' ]]; then
		local new_path=$(prompt_pure_print_path)
		2=${2/'%~'/$new_path}
	fi
	builtin print $@
}
```

sindresorhus:
Would be cool to have a recipe with a screenshot in the wiki though. That way people that want it could add it themselves.

mafredri:
Hehe, cool idea @sindresorhus. The code is basically there anyway :D. Anyone want a stab at it? If not I'll have a look at it soon :).

sindresorhus:
> Thinking a separate recipe section in a new markdown document or some such.

:+1:

PR #149, opened 2015-07-22
Refactor to avoid subshell usage, prefer variables

mafredri:
Performance tests indicated that using a lot of subshells can lead to bad performance as can be seen in [this gist](https://gist.github.com/mafredri/25aa77de809e86d24a63). I didn't want to change everything and hope it made anything better, so I decided to explore what this means in the real world for `pure` and performed some benchmarks.

The script I used for benchmarking can be found: [here](https://gist.github.com/mafredri/7625b223db6c2a1200f4)

#### Results:

To be honest the speed wasn't bad enough for performance improvements to make much of a difference, but at least they are consistently better. Perhaps older systems and maybe even Windows would see better performance from these changes?

##### Branches tested (in order):

* master
* pure-nitro (vcs_info) *[same as PR]*
	* Relies on vcs_info for current git working tree
* pure-nitro (git)
	* Uses `command git rev-parse show-toplevel` for git tree (like master)

##### 10,000 iterations

```
sum(field-1)	max(field-1)	min(field-1)	mean(field-1)	median(field-1)
359.2467112505	0.0578019619	0.0319838524	0.0359246711251	0.03555846215
225.164195295	0.0337250233	0.0200719833	0.0225164195295	0.0222561359
258.7346487026	0.0365149975	0.0230011940	0.0258734648703	0.0255959034
```

**Median improvement:** 13.3 ms (master <> pure-nitro (vcs_info))

##### 1,000 iteration + cpu stress

```
sum(field-1)	max(field-1)	min(field-1)	mean(field-1)	median(field-1)
48.4606709483	0.1176729202	0.0320739746	0.0484606709483	0.0448194742
30.9290690447	0.0984218121	0.0203421116	0.0309290690447	0.0280870199
35.0805423275	0.1091880798	0.0235159397	0.0350805423275	0.0321009159
```

**Median improvement:** 16.7 ms (master <> pure-nitro (vcs_info))

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -26,25 +26,29 @@
 # turns seconds into human readable time
 # 165392 => 1d 21h 56m 32s
 # https://github.com/sindresorhus/pretty-time-zsh
-prompt_pure_human_time() {
-	echo -n " "
-	local tmp=$1
-	local days=$(( tmp / 60 / 60 / 24 ))
-	local hours=$(( tmp / 60 / 60 % 24 ))
-	local minutes=$(( tmp / 60 % 60 ))
-	local seconds=$(( tmp % 60 ))
-	(( $days > 0 )) && echo -n "${days}d "
-	(( $hours > 0 )) && echo -n "${hours}h "
-	(( $minutes > 0 )) && echo -n "${minutes}m "
-	echo "${seconds}s"
+prompt_pure_human_time_to_var() {
+	local human=" " total_seconds=$1 var=$2
+	local days=$(( total_seconds / 60 / 60 / 24 ))
+	local hours=$(( total_seconds / 60 / 60 % 24 ))
+	local minutes=$(( total_seconds / 60 % 60 ))
+	local seconds=$(( total_seconds % 60 ))
+	(( days > 0 )) && human+="${days}d "
+	(( hours > 0 )) && human+="${hours}h "
+	(( minutes > 0 )) && human+="${minutes}m "
+	human+="${seconds}s"
+
+	# store human readable time in variable as specified by caller
+	typeset -g "${var}"="${human}"
 }
 
-# displays the exec time of the last command if set threshold was exceeded
+# stores (into prompt_pure_cmd_exec_time) the exec time of the last command if set threshold was exceeded
 prompt_pure_check_cmd_exec_time() {
-	local stop=$EPOCHSECONDS
-	local start=${prompt_pure_cmd_timestamp:-$stop}
-	integer elapsed=$stop-$start
-	(($elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5})) && prompt_pure_human_time $elapsed
+	integer elapsed
+	(( elapsed = EPOCHSECONDS - ${prompt_pure_cmd_timestamp:-$EPOCHSECONDS} ))
+	prompt_pure_cmd_exec_time=
+	(( elapsed > ${PURE_CMD_MAX_EXEC_TIME:=5} )) && {
+		prompt_pure_human_time_to_var $elapsed "prompt_pure_cmd_exec_time"
+	}
 }
 
 prompt_pure_clear_screen() {
@@ -57,17 +61,26 @@ prompt_pure_clear_screen() {
 }
 
 prompt_pure_check_git_arrows() {
+	# reset git arrows
+	prompt_pure_git_arrows=
+
 	# check if there is an upstream configured for this branch
 	command git rev-parse --abbrev-ref @'{u}' &>/dev/null || return
 
-	local right left arrows
-	right=$(command git rev-list --right-only --count HEAD...@'{u}' 2>/dev/null)
-	left=$(command git rev-list --left-only --count HEAD...@'{u}' 2>/dev/null)
+	local arrow_status
+	# check git left and right arrow_status
+	arrow_status="$(command git rev-list --left-right --count HEAD...@'{u}' 2>/dev/null)"
+	# exit if the command failed
+	(( !$? )) || return
 
-	(( ${right:-0} > 0 )) && arrows="${PURE_GIT_DOWN_ARROW:-⇣}"
+	# left and right are tab-separated, split on tab and store as array
+	arrow_status=(${(ps:\t:)arrow_status})
+	local arrows left=${arrow_status[1]} right=${arrow_status[2]}
+
+	(( ${right:-0} > 0 )) && arrows+="${PURE_GIT_DOWN_ARROW:-⇣}"
 	(( ${left:-0} > 0 )) && arrows+="${PURE_GIT_UP_ARROW:-⇡}"
-	# output the arrows
-	[[ $arrows != "" ]] && echo " ${arrows}"
+
+	[[ -n $arrows ]] && prompt_pure_git_arrows=" ${arrows}"
 }
 
 prompt_pure_preexec() {
@@ -84,8 +97,13 @@ prompt_pure_preexec() {
 }
 
 # string length ignoring ansi escapes
-prompt_pure_string_length() {
-	echo $(( ${#${(S%%)1//(\%([KF1]|)\{*\}|\%[Bbkf])}} ))
+prompt_pure_string_length_to_var() {
+	local str=$1 var=$2 length
+	# perform expansion on str and check length
+	length=$(( ${#${(S%%)str//(\%([KF1]|)\{*\}|\%[Bbkf])}} ))
+
+	# store string length in variable as specified by caller
+	typeset -g "${var}"="${length}"
 }
 
 prompt_pure_preprompt_render() {
@@ -114,13 +132,16 @@ prompt_pure_preprompt_render() {
 		# only redraw if preprompt has changed
 		[[ "${prompt_pure_last_preprompt}" != "${preprompt}" ]] || return
 
-		# calculate length of preprompt for redraw purposes
-		local preprompt_length=$(prompt_pure_string_length $preprompt)
-		local lines=$(( ($preprompt_length - 1) / $COLUMNS + 1 ))
+		# calculate length of preprompt and store it locally in preprompt_length
+		integer preprompt_length
+		prompt_pure_string_length_to_var "${preprompt}" "preprompt_length"
+
+		# calculate number of preprompt lines for redraw purposes
+		integer lines=$(( (preprompt_length - 1) / COLUMNS + 1 ))
 
 		# disable clearing of line if last char of preprompt is last column of terminal
-		local clr="\e[K"
-		(( $COLUMNS * $lines == $preprompt_length )) && clr=""
+		local clr='\e[K'
+		(( COLUMNS * lines == preprompt_length )) && clr=
 
 		# modify previous preprompt
 		print -Pn "\e7\e[${lines}A\e[1G${preprompt}${clr}\e8"
@@ -131,15 +152,15 @@ prompt_pure_preprompt_render() {
 }
 
 prompt_pure_precmd() {
-	# store exec time for when preprompt gets re-rendered
-	prompt_pure_cmd_exec_time=$(prompt_pure_check_cmd_exec_time)
+	# check exec time and store it in a variable
+	prompt_pure_check_cmd_exec_time
 
 	# by making sure that prompt_pure_cmd_timestamp is defined here the async functions are prevented from interfering
 	# with the initial preprompt rendering
 	prompt_pure_cmd_timestamp=
 
 	# check for git arrows
-	prompt_pure_git_arrows=$(prompt_pure_check_git_arrows)
+	prompt_pure_check_git_arrows
 
 	# tell the terminal we are setting the title
 	print -Pn "\e]0;"
@@ -193,11 +214,11 @@ prompt_pure_async_tasks() {
 		prompt_pure_async_init=1
 	}
 
-	# get the current git working tree, empty if not inside a git directory
-	local working_tree="$(command git rev-parse --show-toplevel 2>/dev/null)"
+	# store working_tree without the "x" prefix
+	local working_tree="${vcs_info_msg_1_#x}"
 
 	# check if the working tree changed (prompt_pure_current_working_tree is prefixed by "x")
-	if [[ "${prompt_pure_current_working_tree:-x}" != "x${working_tree}" ]]; then
+	if [[ ${prompt_pure_current_working_tree#x} != $working_tree ]]; then
 		# stop any running async jobs
 		async_flush_jobs "prompt_pure"
 
@@ -210,21 +231,20 @@ prompt_pure_async_tasks() {
 	fi
 
 	# only perform tasks inside git working tree
-	[[ "${working_tree}" != "" ]] || return
+	[[ -n $working_tree ]] || return
 
-	if (( ${PURE_GIT_PULL:-1} )); then
-		# make sure working tree is not $HOME
-		[[ "${working_tree}" != "$HOME" ]] &&
+	# do not preform git fetch if it is disabled or working_tree == HOME
+	if (( ${PURE_GIT_PULL:-1} )) && [[ $working_tree != $HOME ]]; then
 		# tell worker to do a git fetch
-		async_job "prompt_pure" prompt_pure_async_git_fetch "$working_tree"
+		async_job "prompt_pure" prompt_pure_async_git_fetch "${working_tree}"
 	fi
 
 	# if dirty checking is sufficiently fast, tell worker to check it again, or wait for timeout
-	local time_since_last_dirty_check=$(( $EPOCHSECONDS - ${prompt_pure_git_last_dirty_check_timestamp:-0} ))
-	if (( $time_since_last_dirty_check > ${PURE_GIT_DELAY_DIRTY_CHECK:-1800} )); then
+	integer time_since_last_dirty_check=$(( EPOCHSECONDS - ${prompt_pure_git_last_dirty_check_timestamp:-0} ))
+	if (( time_since_last_dirty_check > ${PURE_GIT_DELAY_DIRTY_CHECK:-1800} )); then
 		unset prompt_pure_git_last_dirty_check_timestamp
 		# check check if there is anything to pull
-		async_job "prompt_pure" prompt_pure_async_git_dirty "${PURE_GIT_UNTRACKED_DIRTY:-1}" "$working_tree"
+		async_job "prompt_pure" prompt_pure_async_git_dirty "${PURE_GIT_UNTRACKED_DIRTY:-1}" "${working_tree}"
 	fi
 }
 
@@ -244,7 +264,7 @@ prompt_pure_async_callback() {
 			(( $exec_time > 2 )) && prompt_pure_git_last_dirty_check_timestamp=$EPOCHSECONDS
 			;;
 		prompt_pure_async_git_fetch)
-			prompt_pure_git_arrows=$(prompt_pure_check_git_arrows)
+			prompt_pure_check_git_arrows
 			prompt_pure_preprompt_render
 			;;
 	esac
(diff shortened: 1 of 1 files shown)


mafredri:
Ok, so my results suddenly changed and I realized my branches were faster because there was no upstream configured for them :angry: 

Here's a new run with 1,000 iterations.
```
sum(field-1)	max(field-1)	min(field-1)	mean(field-1)	median(field-1)
36.1003866200	0.0657820702	0.0326840878	0.03610038662	0.0352869034
30.7711567889	0.0422301292	0.0278439522	0.0307711567889	0.02985847
34.5124785908	0.0669791698	0.0309028625	0.0345124785908	0.0334739685
```
Results are even less amazing now :disappointed:

mafredri:
For good measure I did one benchmark in the home-directory as well (no git tasks):

10,000 iterations.
```
sum(field-1)	max(field-1)	min(field-1)	mean(field-1)	median(field-1)
116.8418889043	0.0303239822	0.0107278824	0.01168418889043	0.0115029812
65.9343042389	0.0129218102	0.0060911179	0.00659343042389	0.0065140724
105.8681366418	0.0376369953	0.0095639229	0.01058681366418	0.0102939606
```

pure.zsh
```diff
@@ -210,21 +227,20 @@ prompt_pure_async_tasks() {
...
+	if (( ${PURE_GIT_PULL:-1} )) && [[ $working_tree != $HOME ]]; then
 		# tell worker to do a git fetch
-		async_job "prompt_pure" prompt_pure_async_git_fetch "$working_tree"
+		async_job "prompt_pure" prompt_pure_async_git_fetch "${working_tree}"
 	fi
 
 	# if dirty checking is sufficiently fast, tell worker to check it again, or wait for timeout
-	local time_since_last_dirty_check=$(( $EPOCHSECONDS - ${prompt_pure_git_last_dirty_check_timestamp:-0} ))
-	if (( $time_since_last_dirty_check > ${PURE_GIT_DELAY_DIRTY_CHECK:-1800} )); then
+	integer time_since_last_dirty_check=$(( EPOCHSECONDS - ${prompt_pure_git_last_dirty_check_timestamp:-0} ))
```
sindresorhus: What's the reason for being able to remove the `$` on `EPOCHSECONDS`? Just curious.

pure.zsh
```diff
@@ -57,17 +59,28 @@ prompt_pure_clear_screen() {
...
+	local output
+	output="$(
+		command git rev-list --right-only --count HEAD...@'{u}' 2>/dev/null &&
+		command git rev-list --left-only --count HEAD...@'{u}' 2>/dev/null
+	)"
+	# exit if either command failed
+	(( !$? )) || return
+
+	# split output on newlines and read right / left
+	local arrows right=${${(@f)output}[1]} left=${${(@f)output}[2]}
```
sindresorhus: What's the benefit of doing it like this? Seems a bit convoluted to run the command, combine them, and then split.

pure.zsh
```diff
@@ -27,24 +27,26 @@
...
+	local days=$(( unix / 60 / 60 / 24 ))
+	local hours=$(( unix / 60 / 60 % 24 ))
+	local minutes=$(( unix / 60 % 60 ))
+	local seconds=$(( unix % 60 ))
+	(( days > 0 )) && human+="${days}d "
+	(( hours > 0 )) && human+="${hours}h "
+	(( minutes > 0 )) && human+="${minutes}m "
+	human+="${seconds}s"
+	eval "${var}=\$human"
 }
```
sindresorhus: Would love a PR on https://github.com/sindresorhus/pretty-time-zsh improving it too :heart: :)

pure.zsh
```diff
@@ -57,17 +59,28 @@ prompt_pure_clear_screen() {
...
+
 	# check if there is an upstream configured for this branch
 	command git rev-parse --abbrev-ref @'{u}' &>/dev/null || return
 
-	local right left arrows
-	right=$(command git rev-list --right-only --count HEAD...@'{u}' 2>/dev/null)
-	left=$(command git rev-list --left-only --count HEAD...@'{u}' 2>/dev/null)
+	# check arrow status inside single subshell for performance
+	local output
+	output="$(
```
sindresorhus: I guess you can't declare and assign in one statement?

pure.zsh
```diff
@@ -57,17 +59,28 @@ prompt_pure_clear_screen() {
...
+		command git rev-list --right-only --count HEAD...@'{u}' 2>/dev/null &&
+		command git rev-list --left-only --count HEAD...@'{u}' 2>/dev/null
```
sindresorhus: Can we execute these in parallel, or would they conflict?

pure.zsh
```diff
@@ -57,17 +59,28 @@ prompt_pure_clear_screen() {
 }
 
 prompt_pure_check_git_arrows() {
+	# reset git arrows
+	prompt_pure_git_arrows=
```
sindresorhus: Can we have this function return a value instead of setting a global variable?

pure.zsh
```diff
@@ -268,8 +284,10 @@ prompt_pure_setup() {
 
 	zstyle ':vcs_info:*' enable git
 	zstyle ':vcs_info:*' use-simple true
-	zstyle ':vcs_info:git*' formats ' %b'
-	zstyle ':vcs_info:git*' actionformats ' %b|%a'
+	zstyle ':vcs_info:*' max-exports 2
+	# vcs_info_msg_0_ = brach, vcs_info_msg_1_ = git top level (x prefixed)
```
sindresorhus: I assume you meant: brach => branch

pure.zsh
```diff
@@ -268,8 +284,10 @@ prompt_pure_setup() {
...
+	zstyle ':vcs_info:git*' formats ' %b' 'x%R'
```
sindresorhus: Can you add a comment on what `'x%R'` means?

pure.zsh
```diff
@@ -268,8 +284,10 @@ prompt_pure_setup() {
 
 	zstyle ':vcs_info:*' enable git
 	zstyle ':vcs_info:*' use-simple true
-	zstyle ':vcs_info:git*' formats ' %b'
-	zstyle ':vcs_info:git*' actionformats ' %b|%a'
+	zstyle ':vcs_info:*' max-exports 2
```
sindresorhus: Can you add a comment what this does?

sindresorhus:
:+1: Generally looks good. Added some minor nitpick as inline comments.

-

@zmwangx Interested in reviewing this pull request? Would really appriciate it :)

zmwangx:
Glad to oblige :)

pure.zsh
zmwangx: That's actually the point of this PR: if we return a value, then we need to run this function in a subshell and capture the output with command substitution; however, if we save the result in a global variable, we avoid that subshell, i.e. forking.

pure.zsh
zmwangx: Again, this combines two subshells to one.

pure.zsh
zmwangx: Because all tokens in an arithmetic expression are recursively expanded (as variable names) for values. According to [official doc](http://zsh.sourceforge.net/Doc/Release/Arithmetic-Evaluation.html):

> Named parameters and subscripted arrays can be referenced by name within an arithmetic expression without using the parameter expansion syntax.

Canonical example:

```zsh
> foo=bar
> bar=5
> echo $((foo))
5
```

This is also the case in other shells, e.g., bash, by the way.

zmwangx:
For the `git` left/right thing, I would combine into one command:

```zsh
left_and_right=$(command git rev-list --left-right --count HEAD...@'{u}')
# the separator between two colons is a literal tab
left=$left_and_right[(ws:        :)1]
right=$left_and_right[(ws:        :)2]
```

where the separator here is a literal tab because `--left-right` prints with the format `left<tab>right`.

pure.zsh
sindresorhus: Alright

pure.zsh
sindresorhus: Thanks :)

zmwangx:
I have to say I'm not a big fan of the changes to `prompt_pure_human_time` and `prompt_pure_string_length`... A bit too hackish for my taste... They even allow arbitrary code execution, which isn't really a concern, but just look bad.

Anyway, I don't think there's a better way if we ever want to avoid subshells [^1]. But we can probably make the code somewhat more understandable? The current invocations are plain cryptic without reading the code of the invoked functions [^2]:

```zsh
prompt_pure_human_time prompt_pure_cmd_exec_time $elapsed

prompt_pure_string_length prompt_pure_preprompt_length "${preprompt}"
```

??? [^3]

My suggestion is to rename the functions to `prompt_pure_save_human_time_to_var` and `prompt_pure_save_string_length_to_var`, then invoke like

```zsh
prompt_pure_save_human_time_to_var $elapsed "prompt_pure_cmd_exec_time"

prompt_pure_save_string_length_to_var "$preprompt" "prompt_pure_preprompt_length"
```

More verbose, yes, (they were already pretty verbose anyway) but more understandable at least to me. The quotes around the variable names make them more like names (especially with color coding).

---
[^1]: There are even more hackish ways like temporarily overriding print/echo builtins or playing with file descriptors, cf. http://stackoverflow.com/q/7502981 — definitely not a fan.

[^2]: While we are at it, we should definitely document `prompt_pure_human_time` and `prompt_pure_string_length` precisely. Especially for `string_length`, which we should mention that we are performing prompt substitution on the input string.

[^3]: There's a comment for the latter, but the code itself better be self-documenting.

zmwangx:
I think I'm cool with everything else :+1: 

I can't comment on `vcs_info` related stuff because I've never dealt with this module before.

pure.zsh
zmwangx: Why do you name the input time `unix`? It's a total time measured in seconds, not a POSIX timestamp... I would rather call it `total_seconds` or `seconds_total`.

zmwangx:
I found it rather curious that using `vcs_info` provides a rather noticeable speedup, so I went to look at `vcs_info`'s source code. However, (without looking into details) it seems to be doing exactly the same thing as we do: ([`VCS_INFO_get_data_git#L156`](https://github.com/zsh-users/zsh/blob/57be7b743fbd7afb86edaac3ac86ef34182f86b9/Functions/VCS_Info/Backends/VCS_INFO_get_data_git#L156))

> gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel )

Do you have an explanation for the speedup?

forivall:
My guess would be that vcs_info fetches the data, and then we fetch the data, so the work was being done twice, now it's only being done once. Or, vcs_info caches it somewhere, and only re-runs that command when the current directory changes. But I haven't checked the code, so I'm not sure about my second hypothesis.

pure.zsh
mafredri: @sindresorhus sure I can submit a PR, but I'll leave the eval stuff out, don't think `pretty-time-zsh` would benefit from it :).

@zmwangx good catch, I must've had a mind slip :+1:

pure.zsh
mafredri: You can but declaring and assigning in one statement masks the exit code, example:

```zsh
local output=$(exit 1)
echo $? # 0

output=$(exit 1)
echo $? # 1
```

pure.zsh
mafredri: We could maybe run them as a separate async job, but maybe rewriting the prompt is actually more overhead? I have yet to see this commands perform slowly in any repo.

pure.zsh
mafredri: We could actually, but then we'd have to move the arrows icons somewhere else, e.g. this function returns, `0` for no arrows, `1` for right arrow, `2` for left arrow, and `3` for left+right arrow.

pure.zsh
mafredri: Good eye, thanks!

pure.zsh
mafredri: Absolutely, would be confusing if you didn't look in the correct place otherwise! :)

pure.zsh
mafredri: Yeah, sure.

mafredri:
> For the `git` left/right thing, I would combine into one command:

Good stuff, didn't know of that possibility!

> Do you have an explanation for the speedup?

Yeah, @forivall is pretty much spot on, since `vcs_info` is doing the work anyway, I just removed the duplicate effort.

@zmwangx I'm not a big fan of `eval` either, but can't really think of an alternative, except hard-coding a variable in those functions. I'm not entirely convinced of the naming since they're doing more than just saving to var, but I'll try to think of something since I do agree that atm it can be confusing for anyone reading the code.

zmwangx:
> since `vcs_info` is doing the work anyway, I just removed the duplicate effort.

Okay, then we can probably avoid all the redundant work by ditching `vcs_info` altogether and writing our own git-related calculations? After all, the only things we use from `vcs_info` prior to this PR are the branch and the special action, and I'm not sure if `vcs_info` is doing other unnecessary stuff.

mafredri:
Perhaps `vcs_info` adds a tiny bit of overhead, but as far as I can tell, it doesn't do a whole lot of heavy lifting ([see this gist](https://gist.github.com/mafredri/444debd00a4b41a47bd2)). I like that `vcs_info` would allow us to support other vcs as well, not just git. But as long as we don't maybe it would be best to not use it at all.

zmwangx:
And here are logs of `vcs_info` when run inside a git repo. The logs are generated from the `master` branch and `master~5`, respectively:

https://gist.github.com/zmwangx/a69075c54c75ae757bee
https://gist.github.com/zmwangx/52b4dea8f8a3234e510f

And parsing out the external git commands:

```zsh
> grep -E '\+[A-Za-z_]+:[0-9]+> git ' </tmp/vcs_info_inside_git_repo_master.log
+VCS_INFO_detect_git:9> git rev-parse --is-inside-work-tree
+VCS_INFO_detect_git:10> vcs_comm[gitdir]=+VCS_INFO_detect_git:10> git rev-parse --git-dir
+VCS_INFO_git_getbranch:29> gitbranch=+VCS_INFO_git_getbranch:29> git symbolic-ref HEAD
+VCS_INFO_get_data_git:156> gitbase=+VCS_INFO_get_data_git:156> git rev-parse --show-toplevel

> grep -E '\+[A-Za-z_]+:[0-9]+> git ' </tmp/vcs_info_inside_git_repo_master\~5.log
+VCS_INFO_detect_git:9> git rev-parse --is-inside-work-tree
+VCS_INFO_detect_git:10> vcs_comm[gitdir]=+VCS_INFO_detect_git:10> git rev-parse --git-dir
+VCS_INFO_git_getbranch:29> gitbranch=+VCS_INFO_git_getbranch:29> git symbolic-ref HEAD
+VCS_INFO_git_getbranch:32> gitbranch=+VCS_INFO_git_getbranch:32> git describe --all --exact-match HEAD
+VCS_INFO_get_data_git:156> gitbase=+VCS_INFO_get_data_git:156> git rev-parse --show-toplevel
```

Not bad actually. We can probably cut down a `git rev-parse --is-inside-work-tree` (since the result is already encoded in the return status of `git rev-parse --show-toplevel`), but other than that, not so much. And we would have to maintain all the treatment of special actions, which is not fun. It is certainly possible to do better than `vcs_info` (cf. [Prezto's `git-info`](https://github.com/zmwangx/prezto/blob/master/modules/git/functions/git-info)), but let's not overcomplicate `pure`.

sindresorhus:
> but let's not overcomplicate pure.

:+1:

mafredri:
@zmwangx I realized you can use typeset instead of `eval` for this purpose, can you think of any downsides?

```
❯ typeset -p msg var target
typeset msg=hello
typeset var=target
typeset target=''

❯ typeset $var=$msg

❯ typeset -p msg var target
typeset msg=hello
typeset var=target
typeset target=hello
```

zmwangx:
No objection and can't think of downsides, except that we probably should use

```zsh
typeset -g "$var"="$msg"
```

instead.

* `-g`: without the global flag, `typeset $var=$msg` is equivalent to `local $var=$msg`.
* Quotes: according to currently employed quoting in `pure.zsh`, we seem to operate under the assumption that `SH_WORD_SPLIT` might be set; in that case we need to guard against word splitting.

mafredri:
> No objection and can't think of downsides, except that we probably should use

Yeah, the `-g` I already considered, but quoting is a fair point!

pure.zsh
```diff
@@ -26,25 +26,29 @@
...
+	(( days > 0 )) && human+="${days}d "
+	(( hours > 0 )) && human+="${hours}h "
+	(( minutes > 0 )) && human+="${minutes}m "
+	human+="${seconds}s"
+
+	# store human readable time in variable as specified by caller
+	typeset -g "${var}"="${human}"
 }
 
 # displays the exec time of the last command if set threshold was exceeded
```
zmwangx: This comment is outdated. The exec time isn't displayed, but rather stored in `prompt_pure_cmd_exec_time` and saved for later.

mafredri: You're so hard on me! :)

zmwangx: Well, since comments go outdated without notice all the time, this is a good time to correct some of them :)

mafredri: Just to be clear, I love it that you spot these cockups :+1: ! So thanks!

zmwangx: No problem :smiley_cat:

pure.zsh
```diff
@@ -114,13 +132,15 @@ prompt_pure_preprompt_render() {
 		# only redraw if preprompt has changed
 		[[ "${prompt_pure_last_preprompt}" != "${preprompt}" ]] || return
 
-		# calculate length of preprompt for redraw purposes
-		local preprompt_length=$(prompt_pure_string_length $preprompt)
-		local lines=$(( ($preprompt_length - 1) / $COLUMNS + 1 ))
+		# calculate preprompt length and store it in prompt_pure_preprompt_length
+		prompt_pure_string_length_to_var "${preprompt}" "prompt_pure_preprompt_length"
```
zmwangx: It's me again... Do you have a specific reason for pushing the previously local parameter `preprompt_length` to global scope? As far as I can tell, `prompt_pure_preprompt_length` doesn't serve any global purpose right now. While a global `prompt_pure_last_preprompt_length` might be useful for something else, I don't think it's necessary to change the scoping in this PR.

mafredri: Hmm, how else would you get the value out of the `prompt_pure_string_length_to_var` function? The other alternative would be to move the contents of `prompt_pure_string_length_to_var` inline here and remove that function.

zmwangx: You don't need to do anything special. `typeset -g` in `prompt_pure_string_length_to_var` is capable of modifying local parameters within the caller. According to [doc](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html):

> The `-g` (global) flag is treated specially: it means that any resulting parameter will not be restricted to local scope. Note that this does not necessarily mean that the parameter will be global, as the flag will apply to any existing parameter (even if unset) from an enclosing function.

Example:

```zsh
> function inner { typeset -g outer_local=modified; }
> function outer { local outer_local=original; inner; print $outer_local; }
> outer
modified
```

mafredri: Oh, that's cool. I've read that about `-g` before, but this kind of use case didn't register for me!

sindresorhus:
@mafredri @zmwangx Anything left before merging? :)

zmwangx:
Looks quite ready to me. Probably worth a squash?

mafredri:
Not that I can think of, squashed and ready :+1: !

sindresorhus:
Awesome! Thanks for the review @zmwangx :)

687474703a2f2f74636c686f73742e636f6d2f36763641355a392e676966

zmwangx:
My pleasure :)

PR #153, opened 2015-07-29
Render new preprompt with previous one in mind

### LINKED ISSUE #145: Reduce visual glitches when pre-prompt is same length as the terminal has columns

mafredri:
This fixes #145 by calculating how many lines the previous preprompt assumes in the current terminal window and either erases preprompt lines that are no longer relevant or allocates more space for the new preprompt.

**Note:** In case nr. 2 it would be possible to move the prompt up one line by moving the cursor there and calling `zle .reset-prompt`. However, I believe moving the prompt upwards would be extremely weird behavior for a terminal. This is somewhat of a edge case anyway, and I believe the extra whitespace on top is acceptable.

In action:
pure_preprompt_update

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -21,6 +21,7 @@
 # \e[1G => go to position 1 in terminal
 # \e8   => restore cursor position
 # \e[K  => clears everything after the cursor on the current line
+# \e[2K => clear everything on the current line
 
 
 # turns seconds into human readable time
@@ -133,18 +134,45 @@ prompt_pure_preprompt_render() {
 		[[ "${prompt_pure_last_preprompt}" != "${preprompt}" ]] || return
 
 		# calculate length of preprompt and store it locally in preprompt_length
-		integer preprompt_length
+		integer preprompt_length lines
 		prompt_pure_string_length_to_var "${preprompt}" "preprompt_length"
 
 		# calculate number of preprompt lines for redraw purposes
-		integer lines=$(( (preprompt_length - 1) / COLUMNS + 1 ))
+		(( lines = ( preprompt_length - 1 ) / COLUMNS + 1 ))
+
+		# calculate previous preprompt lines to figure out how the new preprompt should behave
+		integer last_preprompt_length last_lines
+		prompt_pure_string_length_to_var "${prompt_pure_last_preprompt}" "last_preprompt_length"
+		(( last_lines = ( last_preprompt_length - 1 ) / COLUMNS + 1 ))
+
+		# clr_prev_preprompt erases visual artifacts from previous preprompt
+		local clr_prev_preprompt
+		if (( last_lines > lines )); then
+			# move cursor up by last_lines, clear the line and move it down by one line
+			clr_prev_preprompt="\e[${last_lines}A\e[2K\e[1B"
+			while (( last_lines - lines > 1 )); do
+				# clear the line and move cursor down by one
+				clr_prev_preprompt+='\e[2K\e[1B'
+				(( last_lines-- ))
+			done
+
+			# move cursor into correct position for preprompt update
+			clr_prev_preprompt+="\e[${lines}B"
+		# create more space for preprompt if new preprompt has more lines than last
+		elif (( last_lines < lines )); then
+			# move cursor using newlines because ansi cursor movement can't push the cursor beyond the last line
+			printf $'\n'%.0s {1..$(( lines - last_lines ))}
+
+			# redraw the prompt since it has been moved by print
+			zle && zle .reset-prompt
+		fi
 
 		# disable clearing of line if last char of preprompt is last column of terminal
 		local clr='\e[K'
 		(( COLUMNS * lines == preprompt_length )) && clr=
 
 		# modify previous preprompt
-		print -Pn "\e7\e[${lines}A\e[1G${preprompt}${clr}\e8"
+		print -Pn "\e7${clr_prev_preprompt}\e[${lines}A\e[1G${preprompt}${clr}\e8"
 	fi
 
 	# store previous preprompt for comparison


mafredri:
Would it be preferable if instead of pushing the prompt down one line, the preprompt would be truncated in some manner to show that it has been updated while sacrificing some of the "old data", for example:

```
~/Projects/pure preprompt-update-fix
```
becomes
```
~/Projects/pure preprompt-update.. *
```
and
```
~/Projects/pure preprompt-upda.. * ⇣
```

zmwangx:
> Would it be preferable if instead of pushing the prompt down one line, the preprompt would be truncated in some manner to show that it has been updated while sacrificing some of the "old data"

Interesting and nice thought, except that it seems rather nontrivial from an implementation point of view. Were there not a possibility of a change of `git_color`, this would be easy to implement (since `%~` and `${vcs_info_msg_0_}` are not subject to change); just write

```zsh
.. %F{$git_color}${prompt_pure_git_dirty}%f%F{cyan}${prompt_pure_git_arrows}%f$prompt_pure_username%F{yellow}${prompt_pure_cmd_exec_time}%f
```

to the right edge of the terminal, in case of an exceed. However, a change in `git_color` makes it necessary to reconsider the otherwise static `${vcs_info_msg_0_}` chunk. I can't think of anything at the moment except calculating the length of each chunk (or group of chunks) and handle accordingly, which would likely lead to rather complicated code.

Any good ideas? Maybe we can sacrifice a `git_color` change in this extreme case? (Excuse my tendency to make compromises.)

While we're at it, do you find `git_color=red` when git status checking is paused somewhat counterintuitive? Normally when something is paused the UI element is "grayed out". Of course for `pure` the git chunk is already gray... Well, just something to think about.

mafredri:
> Interesting and nice thought, except that it seems rather nontrivial from an implementation point of view.

@zmwangx as usual you have some really great feedback :)! I hadn't actually considered it more than a as a suggestion.

As you pointed out these limitations, the following algorithm came to mind:

```
local last_char
while (( preprompt_length >= wanted_space )); do
	last_char=${preprompt[-1]}
	preprompt=${prepromt[1,$(( ${#preprompt} - 1 ))]}
	prompt_pure_string_length_to_var "${preprompt}" "preprompt_length"
done

preprompt+="${last_char}${extra}"
```

The reasoning here is, once `preprompt_length` becomes smaller than the extra space we're going to take, we know that the last character removed (that decreased the length) was an actual character (not colors, etc), so we add it back at the end.

> Any good ideas? Maybe we can sacrifice a `git_color` change in this extreme case? (Excuse my tendency to make compromises.)

Don't think we need to make any sacrifices, but while we're entertaining this idea, it would be quite interesting to make the pure preprompt only ever take up one line, if it's tight on space it could decrease the path accuracy (e.g. `%2~`, `%3~`, etc..) until it fits.

> While we're at it, do you find `git_color=red` when git status checking is paused somewhat counterintuitive? Normally when something is paused the UI element is "grayed out". Of course for `pure` the git chunk is already gray... Well, just something to think about.

I don't know if it's good or not, it was just the first thing that came to my mind back when I implemented it, that kind of fit the old style. I'm not sure color is the best way to indicate these either, unless we changed the default gray to something else... maybe. But maybe this is worth creating a new issue over for ideas/discussion :)

zmwangx:
> As you pointed out these limitations, the following algorithm came to mind:

I like that, except a technical detail: I'm more inclined to write `${prepromt:0:-1}` or at least `$preprompt[1,-2]`, instead of `${prepromt[1,$(( ${#preprompt} - 1 ))]}`. This solution *looks* painfully slow, but I guess in real world it's negligible (I never benchmarked parameter expansion or anything), and it's stuffed in an async call anyway.

Another solution (as I mentioned above) is to calculate the individual lengths of `%~` and `${vcs_info_msg_0_}` to decide where to cut off, but that's nowhere nearly as extensible, and the code would be ugly.

> it would be quite interesting to make the pure preprompt only ever take up one line

I don't think that's gonna happen, since you can have practically arbitrarily long git branch names ;) I routinely have 15-20 char branch names, and in a tmux window with enough horizontal splits it's just not going to fit along with other stuff. While that is an edge case, the issue addressed in this PR is IMO a rarer edge case...

mafredri:
> I like that, except a technical detail

Yeah, I keep forgetting you can do that :stuck_out_tongue_winking_eye: .

> This solution *looks* painfully slow, but I guess in real world it's negligible (I never benchmarked parameter expansion or anything)

I guess you're referring to calling `prompt_pure_string_length_to_var`? I'm not quite sure, but if we expanded `%~` beforehand (which we anyway have to do if we want to be able to truncate it), there would be nothing left for that function to expand, only strip away the colors, etc. It could also be intelligent and work away n-long chunks at a time, i.e. length is `25`, we want `18`, remove `7`, ok? no. remove `length-18` :D.

> and it's stuffed in an async call anyway.

Actually it's not inside an async call. Only the function run by the worker is truly async, after that it's returned to the shell, and if you do heavy processing afterwards, that would likely lock the users shell for a bit :).

> I don't think that's gonna happen, since you can have practically arbitrarily long git branch names ;)

I don't see how long git branch names has any effect on the suggestion, I mean something like this:

```
~/Projects/pure that
-there-feature*
```
becomes:
```
~/P/pure that-the..*
```
or even:
```
pure that-there-f..*
```

I wouldn't call this an edge case either, as it's a direct effect of having a small terminal window, and I'd imagine a multi-line preprompt is just more distracting and hard to read than it's worth :).

zmwangx:
> It could also be intelligent and work away n-long chunks at a time, i.e. length is 25, we want 18, remove 7, ok? no. remove length-18 :D.

The best strategy might be strip 25-18 the first round, then one at a time. I don't want to over-truncate —precious space will be wasted!

> Actually it's not inside an async call.

Right, stupid me. We can time that to get an idea of how long it will take. My guess is it won't be a problem (even without any preprocessing on our part).

> I don't see how long git branch names has any effect on the suggestion, I mean something like this:

I see. I'm not a big fan of that, because I would lose the ability to grab pwd and branch name directly from the preprompt, which I do reasonably often (iTerm2 makes it very easy to smartly copy stuff on screen). I don't know if others do the same.

Also, how do you deal with maliciously small terminal windows? Say 10 columns wide? ;)

> it's a direct effect of having a small terminal window

I don't know about others, but my terminal windows are 130 columns wide, and split panes are naturally 60+ columns unless I go paranoid and have more than two horizontally (or intentionally shrink one), so multiline prompts are rare for me.

mafredri:
> The best strategy might be strip 25-18 the first round, then one at a time. I don't want to over-truncate —precious space will be wasted!

Isn't that what I did in my example?

> I'm not a big fan of that

The whole point is that you don't, unless you use really, really tiny terminal :P.. The first part lost is the file path, if that can't be truncated further (name of PWD) then you'd lose from git branch, but I digress. These are just ideas.

> so multiline prompts are rare for me

Yeah me too, mine are usually 125 columns wide, and I pretty much never split panes.

zmwangx:
> Isn't that what I did in my example?

I guess I misinterpreted something... Anyway, the point is I agree with you.

> The whole point is that you don't, unless you use really, really tiny terminal :P..

I mean, I would prefer spreading over multiple lines than losing information. But yeah, I guess for folks who routinely use tiny terminals or is routinely ten levels deep in the filesystem, it might be desirable to shrink into one line.

Just for the record, the `sorin` prompt has a [rather intimidating implementation of shrinkage](https://github.com/sorin-ionescu/prezto/blob/f2a826e963f06a204dc0e09c05fc3e5419799f52/modules/prompt/functions/prompt_sorin_setup#L35-L44):

```zsh
```

mafredri:
> Anyway, the point is I agree with you.

:smile:. It's great that we agree, but I don't think this truncating feature will amount to much, it doesn't really fix anything for case nr. 2 (adressed by this PR) without actually forcing pure precmd to always stay on one line.

> I mean, I would prefer spreading over multiple lines than losing information.

Yeah, I think I mostly agree with you, although I would be fine with partial information loss (e.g. really deep directories, etc.)

> Just for the record, the `sorin` prompt has a [rather intimidating implementation of shrinkage](https://github.com/sorin-ionescu/prezto/blob/f2a826e963f06a204dc0e09c05fc3e5419799f52/modules/prompt/functions/prompt_sorin_setup#L35-L44):

Oh wow, that's beyond horrible!

zmwangx:
> but I don't think this truncating feature will amount to much, it doesn't really fix anything for case nr. 2 (adressed by this PR) without actually forcing pure precmd to always stay on one line.

Well, we can probably proceed with the following heuristics for case 2: if a preprompt exceeds whole lines by <=4 characters (or whatever fluctuation we might have), then we truncate it to whole lines.

> Oh wow, that's beyond horrible!

I never really thought about that (nor have I ever attempted to decode that expression), so there might or might not be something better.

mafredri:
> Well, we can probably proceed with the following heuristics for case 2:

I don't know, that kind of seems like arbitrary behavior. As in, it should be quite rare that it happens, and when it suddenly does it's like "wtf is this?", heh.

> I never really thought about that (nor have I ever attempted to decode that expression), so there might or might not be something better.

Not hating on the expression itself, the end result I mean. The paths it produces are completely meaningless to me :D

mafredri:
I'm reading from the lack of recent comments that this is OK for now, let's revisit if it is felt improvements can be made.

PR #158, opened 2015-07-31
Allow customization of the dirty marker

micampe:
We can customize the up/down arrows, this change does the same for the dirty marker.


mafredri:
The reason the arrows are customizable is due to not being able to guarantee the availability of unicode characters. This isn't really a problem with the `*` symbol though. Adding options is easy, but really all we want is to strip them away 😂

Issue #160, opened 2015-08-27
Add option to prevent setting terminal title


alem0lars:
Sometimes you need to **not set** the terminal title automatically.

For example when you use ZSH inside tmux or screen, it's a common practice to name tabs in a comfortable way to remember what's going on on each tab.

Is it possible to introduce a environment variable (or zstyle) to prevent automatic setting of the terminal title?

Thanks :)

mafredri:
Is there a way to detect if the title has been modified or some such? Or is
there a way to tell tmux to ignore the title when you set one manually?
Honestly that's where I think it belongs.

On vacation at the moment, but can take a look when I get back.

alem0lars:
Tmux does nothing particular.. It just picks the terminal title and when you manually set the title it just sets the terminal title to the new value.

I've tried commenting the zsh-add-hook lines and then I am able to manually set the title but ofc it's not a solution since I think it those hooks do other things and sometimes is good to have title automatically set. 

Alessandro Molari

mafredri:
I can't actually reproduce the issue with tmux, after naming the window, it stays named that way:

![](https://www.dropbox.com/s/kus6x6ogk4jwhal/Screenshot%202015-09-09%2013.16.38.png?dl=1)

Are you sure your issue isn't solved by setting `set -g set-titles on` in your tmux.conf ?

alem0lars:
Yes it's not working.
I have that option enabled and also I've tried to manually set it:


![2015-09-09-134517_1440x900_scrot](https://cloud.githubusercontent.com/assets/336957/9763449/15df8a44-56f9-11e5-9994-2753c6ca1b03.png)

![2015-09-09-134651_1440x900_scrot](https://cloud.githubusercontent.com/assets/336957/9763476/3db07d3a-56f9-11e5-9617-9c7557d0da31.png)

![2015-09-09-134531_1440x900_scrot](https://cloud.githubusercontent.com/assets/336957/9763456/1c426d0c-56f9-11e5-929c-6593f3aac128.png)

![2015-09-09-134538_1440x900_scrot](https://cloud.githubusercontent.com/assets/336957/9763458/1f17996c-56f9-11e5-8c0b-824923da0928.png)

----

Instead, when I comment the hooks it works correctly

mafredri:
[...]

PR #164, opened 2015-09-22
Deduplicate code by creating a set_title function

mafredri:
While considering #160 it bothered me a bit that there was duplicated code for setting the title. Lately I've also been thinking about making pure extendable. This PR removes code duplication and allows some form of extendability through hacking around pure.

Example of disabling and restoring title setting functionality:

```zsh
orig_set_title="$(whence -f prompt_pure_set_title)"
pure_restore_set_title() {
	eval "$orig_set_title"
}
pure_disable_set_title() {
	prompt_pure_set_title() {}
}
```

I didn't do this refactor for this effect primarily though, it's just a side-effect.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -84,17 +84,26 @@ prompt_pure_check_git_arrows() {
 	[[ -n $arrows ]] && prompt_pure_git_arrows=" ${arrows}"
 }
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Looks good! :)

PR #175, opened 2015-11-27
Refresh pure screenshot

mafredri:
Tried to mimic the original as much as I could and added a showcase of git dirty while I was at it.


sindresorhus:
Looks good :)

PR #178, opened 2015-12-02
Use builtin cd, in case cd::cd has been redefined

jedahan:
I started to use [enhancd](https://github.com/b4b4r07/enhancd/), which redefined cd::cd to be smarter and more fun to use. However, when in a git directory, [it would hang forever and start writing an infinitely large log file](https://github.com/b4b4r07/enhancd/issues/16). There's probably some changes enhancd could use as well, but this little change fixed the issue for me.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -223,7 +223,7 @@ prompt_pure_async_git_dirty() {
 	local untracked_dirty=$1; shift
 
 	# use cd -q to avoid side effects of changing directory, e.g. chpwd hooks
-	cd -q "$*"
+	builtin cd -q "$*"
 
 	if [[ "$untracked_dirty" == "0" ]]; then
 		command git diff --no-ext-diff --quiet --exit-code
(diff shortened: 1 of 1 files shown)


mafredri:
The `\` prefix is unfamiliar to me, I guess it does the same as the `builtin` command but I've never seen any references to it. I'm not against this change as I don't see any negative impacts from it, but I think perhaps using `builtin cd` would be more descriptive.

sindresorhus:
@mafredri I would prefer `builtin` too, for clarity. Ideally, shouldn't we use `builtin` on any builtin command to ensure we use the pure version?

jedahan:
Yep, I think using `builtin` as much as probably is probably safer and faster

forivall:
That's exactly why i made sure (my custom cd)[https://github.com/forivall/dotfiles/blob/master/plugins/magic-cd/magic-cd.plugin.zsh] still has the default behaviour :)

sindresorhus:
@jedahan Thanks :)

@mafredri Can you make sure we use `builtin` wherever possible?

mafredri:
There are basically two builtins of interest that are called without making sure it's a builtin, `print` and `test`. We could protect the `test` but I don't think `print` (or either, really) makes any sense. It feels premature without having a real reason to do so.

Then there's also `zle`, `integer`, `typeset`, `zmodload`, `autoload`, `add-zsh-hook`, `zstyle`. But protecting these reserved zsh keys / builtins make no sense to me. If a user is overriding them they have a lot bigger problems. Not to mention that there's no way to protect against a `alias builtin="echo nope"` or `alias command="echo nope"` which would mess up pure real hard.

forivall:
I took a quick look at this and was surprised that `builtin` isn't reserved and can be overridden. Although, I can see a situation where someone would want to override `test` or `print` (for testing and printing purposes, obvi)

PR #181, opened 2016-01-13
Added antibody instructions

caarlos0:
[antibody](https://github.com/getantibody/antibody) does basically the same thing as antigen, but it is faster.

Commands are similar, except for the `theme` command that doesn't even exist in antibody.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -152,6 +152,15 @@ antigen bundle mafredri/zsh-async
 antigen bundle sindresorhus/pure
 ```
 
+### [antibody](https://github.com/getantibody/antibody)
+
+Update your `.zshrc` file with the following two lines (order matters):
+
+```
+antibody bundle mafredri/zsh-async
+antibody bundle sindresorhus/pure
+```
+
 ## FAQ
 
 ### My preprompt is missing when I clear the screen with Ctrl+L


mafredri:
I think perhaps the instructions for antigen and antibody could be combined? Feels like the section on integration is getting really long, but maybe that's just me :D.

sindresorhus:
@mafredri That would be kinda confusing since they have different command names, but if you think the readme is getting crowded you could always just move the integration section into a separate file and link to it.

caarlos0:
Thanks!

mafredri:
Nah, It's not that bad yet. Thanks for the PR @caarlos0!

caarlos0:
No problem, thanks @mafredri and @sindresorhus :dancers:

forivall:
Does this mean I should submit a PR for zgen too? it's just `zgen load sindresorhus/pure` and `zgen load mafredri/zsh-async`.

For antigen clones, it should be simple enough to say "Replace `antigen` with `antibody` to use pure with `antibody`." and "Replace `antigen bundle` with `zgen load` to use pure with `zgen`".

Or maybe i'm just butthurt that antibody got mentioned before zgen :fearful:

caarlos0:
@forivall 

> Does this mean I should submit a PR for zgen too? it's just zgen load sindresorhus/pure and zgen load mafredri/zsh-async.
> 
> For antigen clones, it should be simple enough to say "Replace antigen with antibody to use pure with antibody." and "Replace antigen bundle with zgen load to use pure with zgen".

Yeah, I think that makes sense...

> Or maybe i'm just butthurt that antibody got mentioned before zgen  (I have opinions on antigen)

LOL, didn't knew about zgen...

Nice to see that it wasn't just me mad at antigen all this time...

PR #185, opened 2016-01-19
Use append argument of fetch to not clobber FETCH_HEAD

forivall:
I discovered this issue while I was playing with `git subtree`:

```
» git fetch origin v0.5.1 && git rev-parse FETCH_HEAD
From git://github.com/sindresorhus/pure
 * tag               v0.5.1     -> FETCH_HEAD

320e26b8a29f3a4e203820881bfb20c52bc75ffe

» git rev-parse FETCH_HEAD 
f2ca766346345546fecbdeb2f232b3d34f96fc68
```

And with this patch, the bottom SHA will be the same as the top SHA.

The one downside is that, if the user doesn't manually fetch or pull for a long time, the FETCH_HEAD will get quite large. We could manually trim this.

Also, I used the long version of `--append` instead of the short `-a` since `-a` often means all.

Docs: https://git-scm.com/docs/git-fetch#_options

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -239,7 +239,7 @@ prompt_pure_async_git_fetch() {
 	builtin cd -q "$*"
 
 	# set GIT_TERMINAL_PROMPT=0 to disable auth prompting for git fetch (git 2.3+)
-	GIT_TERMINAL_PROMPT=0 command git -c gc.auto=0 fetch
+	GIT_TERMINAL_PROMPT=0 command git -c gc.auto=0 fetch --append
 }
 
 prompt_pure_async_tasks() {


mafredri:
Interesting, could you tell me what relevance this has? I'd like to understand what use case this fixes, is it ever useful to check the FETCH_HEAD anyway?

Does this append one entry during every fetch, even if it's already available in FETCH_HEAD?

forivall:
In my case, I was using FETCH_HEAD with subtree as follows:
```
cd ~; mkdir fork; cd fork; git init
cd ~; git clone upstream-root; cd upstream-root; git subtree split --branch split-subdir --prefix=subdir-that-i-want HEAD
cd ~/fork
git fetch ~/upstream-root split-subdir
git subtree add --prefix=upstream FETCH_HEAD
```
This way, I could add a subtree split from another repo, without pulling in the entire other repo into the fork's repo's history & configuration. (shortly after, i discovered that i could have said `git subtree add --prefix=upstream ~/upstream-root split-subdir`, where subtree uses FETCH_HEAD itself internally. :shrug:)

---

The more common use is to peek at a remote repository's history without configuring the remote, such as

```
git clone git@github.com:sindresorhus/pure && cd pure
git fetch git@github.com:forivall/pure underline-repo-name
git log FETCH_HEAD
git diff HEAD FETCH_HEAD
```
So you can see what the difference is in my fork, but without `git remote add ...`, etc. Most often, FETCH_HEAD is used by scripts, but there's the occasional use case where it's useful in your terminal cli.

mafredri:
Thanks for the detailed explanation. I see now how it can be useful. It does, however, worry me quite a bit that pure will add an entry on every command execution. There's also the fact that one could hypothetically wish to rely on `FETCH_HEAD` as updated by `pure` (as e.g. indicated by an appearing arrow) and in that case the user would be greeted by a stale `FETCH_HEAD`.

sindresorhus:
@mafredri What should we do about this PR?

mafredri:
@sindresorhus I'm not sure. On the one hand this seems like a valid use-case, on the other, I think this PR introduces other issues. I've also tried to find other solutions, but there really doesn't seem to be much leeway with regards to how `git fetch` is implemented.

I could consider merging this PR, but then I'd also want a file-size check to see that FETCH_HEAD doesn't grow too large (and a regular `git fetch` could be performed to trim it down), but maybe that's premature at this point.

sindresorhus:
Then I'd vote we close it. I don't see the risk being worth such an obscure use-case.

forivall:
Well, i guess I'll have to write a PR for git itself then! Hopefully they don't find it obscure (given how many options there are in git anyways...)

torarnv:
I've been hitting this. I was wondering why I always had to do `git fetch foo && git something FETCH HEAD`, instead of doing the two in two separate steps. The fetch done by the prompt will overwrite the stuff I manually fetched it seems.

mafredri:
@torarnv yes, it is unfortunate. As discussed here there is no way around it. It might be possible with low-level git plumbing commands, but we don't have the expertise. If you know of a way to either fix it or work around it we'd welcome the insight and/or PR.

Issue #188, opened 2016-01-21
extraneous characters when used in ansi-term in emacs


randre03:
I'm using Pure inside slimzsh https://github.com/changs/slimzsh as my terminal, but also with ansi-term in emacs. In ansi-term it results in extra characters being returned between prompts. This blog post - http://blog.refu.co/?p=1277 - describes exactly the issue as well as a solution for another zsh framework. I reached out to slimzsh's creator @changs for a fix, but after some assistance he indicated this issue would need to be resolved in the code for Pure.

I tried something along the lines of the suggested workaround in the provided blog post wherein I enclosed the "prompt_pure_set_title" function inside a conditional to check if it's running in emacs terminal, but that did not work. Any thoughts? Thanks

mafredri:
Since I don't know how to reproduce this (I don't use emacs), let's start slow. Does it work if you add this to your `.zshrc` (after `prompt pure`)?

```zsh
# Disable set title
prompt_pure_set_title() {}
```

randre03:
Heavens to Betsy! That  worked. Thanks chap!

mafredri:
Gald it works for you! Keep in mind this disables the set title altogether, if you still want to keep it in other terminals then put it inside a conditional statement that can properly determine the terminal being used :+1: .

PR #190, opened 2016-01-21
Add PURE_SET_TITLE option

changs:
I propose to add an option to be able to decide whether Pure should change a terminal title.
In some cases like https://github.com/sindresorhus/pure/issues/188 one may want to be able to disable this feature.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -85,7 +85,7 @@ prompt_pure_check_git_arrows() {
 }
 
 prompt_pure_set_title() {
-	# do not set title if PURE_SET_TITLE is equal to 0
+	# do not set title if PURE_SET_TITLE is set to 0
 	(( ${PURE_SET_TITLE:-1} == 0 )) && return
 
 	# tell the terminal we are setting the title
--- a/pure.zsh
+++ b/pure.zsh
@@ -85,6 +85,9 @@ prompt_pure_check_git_arrows() {
 }
 
 prompt_pure_set_title() {
+	# do not set title if PURE_SET_TITLE is equal to 0
+	(( ${PURE_SET_TITLE:-1} == 0 )) && return
+
 	# tell the terminal we are setting the title
 	print -n '\e]0;'
 	# show hostname if connected through ssh
--- a/readme.md
+++ b/readme.md
@@ -109,6 +109,10 @@ Defines the git down arrow symbol. The default value is `⇣`.
 
 Defines the git up arrow symbol. The default value is `⇡`.
 
+## `PURE_SET_TITLE`
+
+Set `PURE_SET_TITLE=0` to prevent Pure from setting terminal title.
+
 ## Example
 
 ```sh


sindresorhus:
We already solved that issue. Do you need this for something? What?

changs:
I'm not sure if overriding function is a good solution. I'd rather do something like:
``` 
if [[ -n ${EMACS} ]]; then
    PURE_SET_TITLE=0
fi
```
than
```
if [[ -n ${EMACS} ]]; then
    prompt_pure_set_title() {} 
fi
```
in my config. 
Besides Emacs, I sometimes want to have my custom function to set the title.
I agree though, that my usage may be an edge case not relevant enough to add it to Pure :)

sindresorhus:
What if we put `if [[ -n ${EMACS} ]]; then
    prompt_pure_set_title() {} 
fi` in pure itself?

changs:
I think that this is a little bit too specific since there are probably other terminals that don't support it and also other use cases. From my perspective it'd be nice to just has the ability to somehow disable `prompt_pure_set_title`.

sindresorhus:
From my perspective, it's much nicer when something *just works*. Adding options has a tendency to add up, and we'd really like Pure to stay simple.

Here's what happens when you just keep adding options:

horrible-ui

sindresorhus:
@mafredri Thoughts?

mafredri:
I'm not quite sure. I would like to avoid adding the option, but I would argue that overriding a function isn't really simpler either (almost the same as an option, if documented). Oh-my-zsh seems to take into account different kinds of terminals [here](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/termsupport.zsh), but they also provide an option to disable settings the title.

There's also #160 that expressed a wish to disable settings the title, although motivations remained unclear.

I think we could easily maintain the list of `TERM`s (and we probably should), but I'm not sure it's the right choice to deny our users the option to disable setting the title.

mafredri:
Could one option be to split out the settings of the title into a new project, e.g. `pure-title`? That would give users the option to only load it when relevant, and for us to maintain special title use-cases separately from pure. I think it could make sense to think of them as separate projects. `pure` is the visual layout of the prompt, whereas `pure-title` is more of a terminal feature.

supercrabtree:
+1 for splitting it out to a new plugin. The default settings for how pure modifies the title doesn't work for me at all. And I think @mafredri has a good point, the prompt and the title are two separate things.

sindresorhus:
I don't like the idea of splitting. The terminal title is part of the prompt. We provide a nice opinionated experience that works for the 95%. The rest can just modify it the way they want. It's impossible to please everyone.

sindresorhus:
> The default settings for how pure modifies the title doesn't work for me at all.

Why?

supercrabtree:
When I say it doesn't work for me, I just mean that it doesn't help me in my workflow understand what is going on in that window, not that it is actually broken. But it's just to do with my setup and the way I use tabs and panes in iterm, and I am more than happy just modifying. It's not a strong +1, I was saying I thought it was a good idea.

mafredri:
Alright, I kind of liked the idea of separating set-title, but I don't feel strongly about it, I'm fine with keeping it in `pure`.

What route would you like to take with this @sindresorhus, merge this PR or add an exception for emacs? Personally I'm kind of leaning towards an `(( ${+EMACS} ))`-kind of solution, and in that case I could add an exception for #191 as well (although that solution should work universally and could be made default and thus skip the usage of two ansi codes).

mafredri:
Just thought of another possibility, (I haven't investigated yet if it's possible) to check `$TERMCAP` for terminal capability to decide if it can do the required operations. But as said, I have not yet investigated if it's possible in this case or not. If anyone knows more about what's available in TERMCAP, I'm all ears :).

sindresorhus:
>  add an exception for emacs? Personally I'm kind of leaning towards an (( ${+EMACS} ))-kind of solution, and in that case I could add an exception for #191 as well (although that solution should work universally and could be made default and thus skip the usage of two ansi codes).

:+1: 

> Just thought of another possibility (I haven't investigated yet if it's possible), to check $TERMCAP for terminal capability to decide if it can do the required operations. But as said, I have not yet investigated if it's possible in this case or not. If anyone knows more about what's available in TERMCAP, I'm all ears :).

@qix- might know.

Qix-:
> check $TERMCAP for terminal capability to decide if it can do the required operations

Just a nit; `$TERMCAP` doesn't mean anything. It's `$TERM` you're looking for ;)

Secondly, as far as I can tell (I went back and double checked) terminal title isn't something covered by the Termcap database. Terminals (not emulators, but _real_ terminals) didn't have windows or any special GUI like that. No title bar, nothing. Termcap didn't need to know about titles because it doesn't pertain to TTYs themselves.

mafredri:
> Just a nit; `$TERMCAP` doesn't mean anything. It's `$TERM` you're looking for ;)

Actually it was `termcap` that I was interested in, `$TERM` only gives the name, e.g. `xterm`, `eterm`, etc. I was hoping for a way to check for this certain capability e.g. out of `termcap` :P.

> Secondly, as far as I can tell (I went back and double checked) terminal title isn't something covered by the Termcap database. Terminals (not emulators, but _real_ terminals) didn't have windows or any special GUI like that. No title bar, nothing. Termcap didn't need to know about titles because it doesn't pertain to TTYs themselves.

Thanks, appreciate the info! :)

Issue #191, opened 2016-01-29
emacs shell 


Fetz:
weird 7 and 8 symbols in emacs shell (spacemacs)

<img width="415" alt="screen shot 2016-01-29 at 11 43 13" src="https://cloud.githubusercontent.com/assets/1071113/12674796/87de1d2e-c67d-11e5-8ff9-b5347522f247.png">

mafredri:
Does the solution in #188 help?

Fetz:
I put in .zshrc after pure but still doesn't work

```
prompt pure
prompt_pure_set_title() {}
```

I notice it only happens when pure updates git status

mafredri:
Ok. Since I see the 7 at the beginning and 8 at the end I suspect your terminal isn't supporting saving the current cursor position:

```
# \e7   => save cursor position
# \e8   => restore cursor position
```

I believe the offending line is in [pure.zsh#L187](https://github.com/sindresorhus/pure/blob/master/pure.zsh#L187).

Without these it is impossible to update the prompt and return the cursor to it's original position :/

mafredri:
Actually I lied about it not being possible to return the cursor to it's original position.

This change should work for you (at least as a band-aid, assuming the rest of the ansi codes are supported), not sure what other repercussions it could have. There are lots of states the terminal could be in that should be taken into consideration.

```diff
diff --git a/pure.zsh b/pure.zsh
index a061851..6b1a10c 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -184,7 +184,8 @@ prompt_pure_preprompt_render() {
                (( COLUMNS * lines == preprompt_length )) && clr=

                # modify previous preprompt
-               print -Pn "\e7${clr_prev_preprompt}\e[${lines}A\e[1G${preprompt}${clr}\e8"
+               print -Pn "${clr_prev_preprompt}\e[${lines}A\e[1G${preprompt}${clr}\n"
+               zle && zle .reset-prompt
        fi

        # store previous preprompt for comparison
```

sindresorhus:
This prompt is for ZSH, not Emacs shell.

mafredri:
@sindresorhus The problem is that you can run ZSH inside emacs, I guess :P

sindresorhus:
@mafredri And people are free to do that, but anything other than running ZSH is IMHO unsupported from our perspective. It's hard enough maintaining something that works fine on just vanilla ZSH.

mafredri:
[...]

PR #196, opened 2016-02-15
Emacs fixes for garbled terminal output

mafredri:
This PR fixes the issues raised in #188 and #191. It also takes precedence over #190 in favor of not introducing new configuration options.

Dependency on ansi codes `\e7` (save cursor position) and `\e8` (restore cursor position) has been removed by introducing a call to `zle reset-prompt` instead, I have not found any negative side-effects to this in my testing.


sindresorhus:
Been using this since yesterday and seems to be working fine :)

Issue #189, opened 2016-01-21
Git is slow on sshfs: add option to disable git-prompt for sshfs mounted folders


Hotschke:
https://github.com/nojhan/liquidprompt#known-limitations-and-bugs (`LP_DISABLED_VCS_PATH`) and https://github.com/arialdomartini/oh-my-git#disabling-oh-my-git offer each a possibility to disable the git-prompt features for certain folders/git repos.

In the spirit of a pretty, minimal and fast ZSH prompt, I would actually expect such a possibility. Otherwise this prompt is not an option for me since git is very slow via sshfs and makes your slim prompt unusable slow.

I hope you find it worthwhile adding a way to disable git feature in certain situations.

mafredri:
Does commenting [this line](https://github.com/sindresorhus/pure/blob/master/pure.zsh#L209) help?

Out of interest, would you mind telling me the output of the following commands (the timings) over sshfs:

```zsh
time command git rev-parse --abbrev-ref @'{u}'
time command git rev-list --left-right --count HEAD...@'{u}'
time command git rev-parse --show-toplevel
```

The two first commands are the only ones called explicitly by pure, all other git commands are either run by vcs_info or asynchronously.

Hotschke:
Thanks for your feedback. I appreciate your efforts.

Commenting your suggested line would disable git additions to the prompt at all. Is this correct?

The result of your commands:

    command git rev-parse --abbrev-ref @'{u}'  0,00s user 0,00s system 1% cpu 0,332 total
    command git rev-list --left-right --count HEAD...@'{u}'  0,00s user 0,00s system 0% cpu 0,656 total
    command git rev-parse --show-toplevel  0,00s user 0,00s system 2% cpu 0,183 total

I acces my home folder on the university server from my macbook to edit different projects which are kept under version control. It is a very convenient and seamless method of working from my macbook. I don't need to sync anything.  I can spontaneously edit a different file without having to configure something. I can use my own version of tex compiler, macvim, the pdf viewer skim, dash.app (all the osx gui apps I don't have on the linux machine in the university, only on my laptop).
The home folder in the university is professionally backuped. It is a safe place to keep my files. If my macbook hard drive crashes, I don't lose anything.
To cut the story short, I currently use liquidprompt which does what I want, but offers to many features I don't use.

I just thought that there are many people who like sshfs and would consider using your zsh prompt as well. If you don't use sshfs and don't see any gain in sshfs, I can understand that your not interested in this.

[...]

PR #200, opened 2016-03-03
Don't check git info if in a fuse directory

sevanteri:
Not sure if I implemented this in the right place or the right way. My zsh skills aren't that great.

But yeah, checking git info in a fuse directory is mostly not ideal as it might take quite a lot of time, for example with sshfs.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -202,14 +202,19 @@ prompt_pure_precmd() {
 	# with the initial preprompt rendering
 	prompt_pure_cmd_timestamp=
 
-	# check for git arrows
-	prompt_pure_check_git_arrows
+	# check if we are in a fuse directory
+	if [[ `stat -f -L -c %T $PWD` == *fuseblk* ]]; then
+		unset prompt_pure_git_arrows
+		unset vcs_info_msg_0_
+	else
+		# check for git arrows
+		prompt_pure_check_git_arrows
+		# get vcs info
+		vcs_info
+	fi
 
 	# shows the full path in the title
 	prompt_pure_set_title 'expand-prompt' '%~'
-
-	# get vcs info
-	vcs_info
 
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks


forivall:
IMO, this is something that we should consider here. However, since NTFS is also mounted by fuse, do you know if there's a way to check that as well? The only thing I can think of is to check if the mount point points to an actual device or not.

Also, this will need to be made compatible with OSX.

sevanteri:
Yea, this reeeaaally needs more thought put into it. I didn't even realise to take OSX into consideration...

forivall:
I mean, for OSX compatibility, as long as it doesn't break, it should be fine, and then the people here who do use OSX can get it working if needed.

mafredri:
There is actually already a [solution](https://github.com/sindresorhus/pure/issues/189#issuecomment-173844254) to this problem (albeit a slightly different approach) as discovered by @Hotschke. Does it suffice (for now?) or is this something we need? If yes, there's at least no need to do the stat check every time, only on directory change (we already use a hook for that). Also, the [zsh/stat](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fstat-Module) module could potentially be used for better cross-platform compatibility.

An idea I've been entertaining is to simplify pure (e.g. removing `PURE_GIT_PULL`, `PURE_GIT_UNTRACKED_DIRTY` and `PURE_GIT_DELAY_DIRTY_CHECK`) and replace them with a simple way of disabling pure git features on a per-folder basis (relying under the hood on the zstyle/vcs_repo syntax).

E.g. something like an associative array (this could also be made non-associative is completely disabling git features instead of being given the choice of a subset):

```zsh
PURE_FOLDER_GIT_FEATURES=(
    /home none
    /alt/path(|/*) all
    /third/path(|/*) no-pull
)
```

This is all still very much in the air and I'd like some more input on it, maybe I should open a new issue about it.

sevanteri:
Yes, a new issue would a better place to work on this.

norpol:
You could check if the current directory is a fuse-mount by something like `df -T $PWD | awk '{ print $2 }' | grep -i "fuse" -q`.

PR #201, opened 2016-03-17
Update and simplify oh-my-zsh integration instructions

### LINKED ISSUE #194: oh-my-zsh integration fix
Hello!

**Thank you for all your works!**

Just to say I followed the oh-my-zsh integration guide but I had to add a `themes` directory in the `custom` directory and then change the location for the symlinks:

```
Symlink (or copy) pure.zsh to ~/.oh-my-zsh/custom/themes/pure.zsh-theme
Symlink (or copy) async.zsh to ~/.oh-my-zsh/custom/themes/async.zsh
```

Without this change I got 2 errors when I started the terminal:
- `too many symlinks in async.zsh`
- `no theme "pure" in oh-my-zsh themes`

Now everything works correctly.

Also I didn't have to add the following:

```
autoload -U promptinit && promptinit
prompt pure
```

Maybe we can add to the oh-my-zsh section that these 2 lines are not needed.

Bye!

fvargas:
As per [Oh My Zsh: Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes), the theme should go in the `.oh-my-zsh/custom/themes/` directory. Doing so will allow it to override the included *pure* theme without having to delete it.

Additionally, there is no need to symlink or copy `async.zsh` into `.oh-my-zsh/custom/` since it should already exist in `$fpath` given a proper installation.

Resolves #194 (looks like others have encountered this too)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -134,10 +134,8 @@ To have commands colorized as seen in the screenshot install [zsh-syntax-highlig
 
 ### [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh)
 
-1. Remove competing theme included in oh-my-zsh `~/.oh-my-zsh/themes/pure.zsh-theme`
-2. Symlink (or copy) `pure.zsh` to `~/.oh-my-zsh/custom/pure.zsh-theme`
-3. Symlink (or copy) `async.zsh` to `~/.oh-my-zsh/custom/async.zsh`
-4. Add `ZSH_THEME="pure"` to your `.zshrc` file.
+1. Symlink (or copy) `pure.zsh` to `~/.oh-my-zsh/custom/themes/pure.zsh-theme`.
+2. Set `ZSH_THEME="pure"` in your `.zshrc` file.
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Did you successfully try this recipe?

LGTM. @mafredri ?

mafredri:
I haven't tried it out but other than that, looks good to me too!

fvargas:
I've been using it with this configuration and everything seems to work correctly.

I specifically verified the following:
- It does override the *pure* theme included by Oh-My-Zsh.
- Checking for local and remote branch changes does occur and it shows the arrows accordingly (without the inclusion of `async.zsh` in the `custom/` directory.
  - Unless the inclusion of `async.zsh` in `custom/` was meant to override some other `async.zsh` used by Oh-My-Zsh, this should be fine.

Thanks for a fantastic theme! :sparkles:

sindresorhus:
Alright. Thank you. Glad to hear you enjoy it :)

PR #202, opened 2016-03-20
Fix theme name under `Tips'

fvargas:
Fix the theme name to correctly correspond to the theme used in the
screenshot.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -125,7 +125,7 @@ prompt pure
 
 ## Tips
 
-[Tomorrow Night Eighties](https://github.com/chriskempson/tomorrow-theme) theme with the [Droid Sans Mono](http://www.google.com/webfonts/specimen/Droid+Sans+Mono) font (15pt) is a beautiful combination, as seen in the screenshot above. Just make sure you have anti-aliasing enabled in your Terminal.
+[Tomorrow Night](https://github.com/chriskempson/tomorrow-theme) theme with the [Droid Sans Mono](http://www.google.com/webfonts/specimen/Droid+Sans+Mono) font (15pt) is a beautiful combination, as seen in the screenshot above. Just make sure you have anti-aliasing enabled in your Terminal.
 
 To have commands colorized as seen in the screenshot install [zsh-syntax-highlighting](https://github.com/zsh-users/zsh-syntax-highlighting).
 


sindresorhus:
Hmm, what makes you think that?

fvargas:
The background color used by Tomorrow Night Eighties is lighter than the color seen in the screenshot. Whereas the background color used by Tomorrow Night matches the color in the screenshot.

fvargas:
To further clarify, it seems to me the screenshots shown in [tomorrow-theme](https://github.com/chriskempson/tomorrow-theme#take-a-look) have the correct colors corresponding (at least) to the provided iTerm2 theme files. However, the colors described in the [Hex Palettes](https://github.com/chriskempson/tomorrow-theme#hex-palettes) section don't correspond to their respective themes. The background color seen in your screenshot is #2d2d2d, which, according to the Hex Palettes, does correspond to Tomorrow Night Eighties. However, if we look at the actual theme files in that repo or the screenshots in that same readme, we see that it does not correspond. Thus, my assertion remains that the theme used in your screenshot most closely corresponds to the current Tomorrow Night theme.

My explanation? When @chriskempson created the base-16 themes, he took what was at that point the Tomorrow Night theme to create the base-16-Tomorrow theme. He then took the Tomorrow Night Eighties theme and used it to create a new (and slightly different) Tomorrow Night theme and additionally created a new, lighter-colored Tomorrow Night Eighties theme. This change was reflected in the repo files and in the screenshots in the readme, but not in the Hex Palettes section. Therefore, the theme you were using at the time you took your screenshot was in fact Tomorrow Night Eighties. However, that Tomorrow Night Eighties theme later (about 4 years ago?) became the Tomorrow Night theme that is presently available in the repo (subject to minor alterations).

I currently use the base-16-Tomorrow theme, so I've poked around these repos a bit. This is just my conjecture, but it seems to explain everything. When making this PR, I was under the assumption it was just a simple mix up; I didn't realize there was so much more to it until just now :P

mafredri:
I'm the one who updated the screenshot last, and it might be that I either messed up with the color scheme (I did try to use the same ones as the original) or that the color is affected by window transparency. Either way, this doesn't change @sindresorhus recommendation of using the *Tomorrow Night Eighties* theme :).

fvargas:
I suppose the point I'm making is just that the recommendation is not consistent with the screenshot it references. That said, I was also misunderstanding the problem earlier.

If you look at [tomorrow-theme](https://github.com/chriskempson/tomorrow-theme), you'll notice the screenshots for *Tomorrow Night Eighties* have lighter colors than the screenshot in this repo. That is because the colors for *Tomorrow Night Eighties* are specified in the Generic RGB color space as opposed to sRGB. The screenshot in this repo shows the *Tomorrow Night Eighties* theme using the sRGB color space. The interpretation of Generic RGB colors as sRGB generally results in darker colors, as seen in this case. This is also the reason why the screenshot appeared off to me in the first place.

The colors have been off since before the most recent screenshot, so it sounds to me like the fix is just to mention that it's interpreting the colors as sRGB rather than Generic RGB.

I realize this is a fairly minor detail, but here we are ... :smile:

Further reading: https://github.com/chriskempson/tomorrow-theme/issues/269

mafredri:
@fvargas interesting insights, and I'm impressed by the amount of effort you put into this detective work :clap:. If you would like to update the readme to note about the RGB differences, I'd be happy to accept the PR :+1:.

PR #216, opened 2016-07-13
Correct link to Droid Sans Mono Font

danez:


mafredri:
Thanks, totally wouldn't have spotted this :)!

PR #219, opened 2016-07-29
Fix link to bash port

Yeti-or:


readme.md
```diff
@@ -195,7 +195,7 @@ On a default setup, running the command `kldload pty` should do the trick. If yo
 ## Ports
 
 * **Bash**
-	* [sapegin/dotfiles](https://github.com/sapegin/dotfiles)’s [prompt](https://github.com/sapegin/dotfiles/blob/master/includes/bash_prompt.bash) and [color theme](https://github.com/sapegin/dotfiles/tree/master/color) for `Terminal.app`.
+	* [sapegin/dotfiles](https://github.com/sapegin/dotfiles)’s [prompt](https://github.com/sapegin/dotfiles/blob/bash/includes/bash_prompt.bash) and [color theme](https://github.com/sapegin/dotfiles/tree/master/color) for `Terminal.app`.
```
mafredri: Would you mind changing this to link directly to [the blob](https://github.com/sapegin/dotfiles/blob/dd063f9c30de7d2234e8accdb5272a5cc0a3388b/includes/bash_prompt.bash) instead of a branch? Should be more reliable than linking to a branch :)

Yeti-or: @sapegin are you planning to remove branch `bash`?

sapegin: No, but I’m not going to update it either.

PR #222, opened 2016-07-30
Add my bash port

anirudhb:
Hi @sindresorhus. I finished my bash port, and I would like to add it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -196,7 +196,7 @@ On a default setup, running the command `kldload pty` should do the trick. If yo
 
 * **Bash**
 	* [sapegin/dotfiles](https://github.com/sapegin/dotfiles)’s [prompt](https://github.com/sapegin/dotfiles/blob/dd063f9c30de7d2234e8accdb5272a5cc0a3388b/includes/bash_prompt.bash) and [color theme](https://github.com/sapegin/dotfiles/tree/master/color) for `Terminal.app`.
-	* [anirudhb/purer](https://github.com/anirudhb/purer): Synchronous Git, NPM, and SSH check, shows outdated dependencies, able to resolve to relative path, and username is always shown. With color!
+	* [anirudhb/purer](https://github.com/anirudhb/purer): Synchronous Git, npm, and SSH check, shows outdated dependencies, able to resolve to relative path, and username is always shown. Uses chalk for color
 * **Fish**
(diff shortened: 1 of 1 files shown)


sindresorhus:
Include a screenshot in your readme so people can see how it looks like.

I would also recommend publishing the prompt to npm so people can easily install it, like Pure.

readme.md
```diff
@@ -196,6 +196,7 @@ On a default setup, running the command `kldload pty` should do the trick. If yo
 
 * **Bash**
 	* [sapegin/dotfiles](https://github.com/sapegin/dotfiles)’s [prompt](https://github.com/sapegin/dotfiles/blob/dd063f9c30de7d2234e8accdb5272a5cc0a3388b/includes/bash_prompt.bash) and [color theme](https://github.com/sapegin/dotfiles/tree/master/color) for `Terminal.app`.
+	* [anirudhb/purer](https://github.com/anirudhb/purer): Synchronous Git, NPM, and SSH check, shows outdated dependencies, able to resolve to relative path, and username is always shown. With color!
```
sindresorhus: NPM => npm

-

> able to resolve to relative path

What does this even mean?

> With color!

I don't see how this is a selling point? All prompts have color.

anirudhb:
As for npm, should I use postinstall?

sindresorhus:
@anirudhb Yes, see Pure's package.json, but needs to do something that is Bash compatible.

anirudhb:
Relative path means: you can use `set-relative-path` to make it look like (in the prompt) the set path is the root directory. Use `reset-relative-path` to reset it. IMO, this is easier when dealing with long paths.

anirudhb:
Also, on Ubuntu, the default prompt has **no color**.

anirudhb:
Done, merge?

sindresorhus:
Hmm, not sure about it being a Pure port. It looks quite different. You called it `purer`, but it's actually a lot less pure. It's not clear what `[GIT NPM]` means (They should also be correctly cased). The `0 outdated package(s)` makes it look ugly. That text should not be shown when there are no outdated packages and should IMHO not be red.

anirudhb:
[GIT NPM] is what I called 'tags' in my README. Also, I did this in one day, so there's a lot of room for improvement.

sindresorhus:
Ok, it has potential, but still needs more work. I'm gonna close this now, but happy to reconsider when you've worked on it some more and it has had time to mature :)

PR #227, opened 2016-08-31
Use ; instead of && for promptinit

### LINKED ISSUE #226: Semi-colon should be used when autoloading promptinit
In regards to `promptinit`, why use `&&`? There is no return value to check; a semi-colon would suffice.

https://github.com/zsh-users/zsh/blob/master/Functions/Prompts/promptinit#L5

leoj3n:
Closes #226


mafredri:
Looks good, thanks! =)

PR #247, opened 2016-11-08
Mention intelfx/pure fork in readme

intelfx:
Recently I've forked this prompt theme for better (more async) Git integration and [Solarized](http://ethanschoonover.com/solarized)-friendly colors. The fork is [intelfx/pure](https://github.com/intelfx/pure).

I've rewritten much of the code in one big commit so you're probably not going to merge that back, but maybe you're interested in mentioning my fork along with others.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -201,6 +201,7 @@ On a default setup, running the command `kldload pty` should do the trick. If yo
 	* [rafaelrinaldi/pure](https://github.com/rafaelrinaldi/pure), support for bare Fish and various framework ([Oh-My-Fish](https://github.com//oh-my-fish/oh-my-fish), [Fisherman](https://github.com//fisherman/fisherman) and [Wahoo](https://github.com//bucaran/wahoo)).
 * **Zsh**
   * [therealklanni/purity](https://github.com/therealklanni/purity): a more compact current working directory, important details on the main prompt line, and extra Git indicators.
+  * [intelfx/pure](https://github.com/intelfx/pure): Solarized-friendly colors, highly verbose and fully async Git integration
 
 ## Team
 


sindresorhus:
Nice! I'm curious, how is your Git integration more async? And what does Solarized-friendly colors mean? We just use the default color set which adheres to the terminal theme.

mafredri:
I'm glad you've done the work for this as it's a path I've been entertaining for pure as well. One reason I never did was that I was worried it would actually make pure feel more slow due to the delay in receiving the info from the async worker.

I did try out your fork and it does seem slightly more responsive than pure when the cache is primed, so to speak (always slightly faster to render prompt though). However, I do notice the slight delay in e.g. switching branches. It's impressive work, and maybe something we should consider for pure as well, I'm not sure, good job on it anyway.

One thing that strikes me in the eye though is when entering a git repository, the prompt seems really busy and does like a million updates (well, likely it's something like 3, but still). I replaced git with a `sleep 2; _git $@` and saw the issue:

```
~/Projects/pure mafredri@Franky
~/Projects/pure master?w ?u mafredri@Franky
~/Projects/pure master ?u mafredri@Franky
~/Projects/pure master mafredri@Franky
```

Basically the placeholders causes a lot of stress on my eyes due to their rapid removal. Also, it seems like the branch doesn't get "cached" before the 3rd prompt after entering a directory (first prompt is as above, second blinks the branch name and third+ does no updates).

PS. Would you mind adding that you always display `username@Host` as well? I find this a pretty big change from pure that actually makes this fork *less* pure.

intelfx:
@sindresorhus 

> I'm curious, how is your Git integration more async?

In your work, only worktree checks (`git diff`) and upstream fetch (`git fetch`) are asynchronous. Branch name resolution (`vcs_info`) and upstream checks (`git rev-list --left --right`) are performed synchronously, yet they can take >1 minute each on big repositories and cold VFS caches (and even with warm cache the branch name resolution introduces a perceptible delay).

In my fork, all Git operations are performed in background. This incurs more overhead, but the prompt feels snappier.

> And what does Solarized-friendly colors mean? We just use the default color set which adheres to the terminal theme.

Solarized is a color palette.

Solarized palette

It works by replacing the terminal palette for both regular (0-7) and bold (8-15) color indices. The problem is that in Solarized bold colors are *not* similar (in hue) to their non-bold counterparts (to be precise: bolds are used for monotones, regulars are used for accent colors). Consequently, when your color choices are paired with the Solarized palette, certain pieces of the prompt become illegible.

I have tweaked the indices to make it work with Solarized.

intelfx:
@mafredri 

> However, I do notice the slight delay in e.g. switching branches.

How exactly?

> the prompt seems really busy and does like a million updates (well, likely it's something like 3, but still).
> Basically the placeholders causes a lot of stress on my eyes due to their rapid removal.

Yes, the placeholders (`?w` and `?u`) are not particularly beautiful, but I wanted to highlight that the data is not yet ready rather than just show nothing and then let the `*` and arrows appear out of nowhere.

> Also, it seems like the branch doesn't get "cached" before the 3rd prompt after entering a directory (first prompt is as above, second blinks the branch name and third+ does no updates).

Could you please elaborate? The branch name is refreshed each time, that's true, but the old value should be cached until the new one appears, so we always render the prompt with _some_ branch name in it.

> PS. Would you mind adding that your fork always displays username@Host as well? I find this a pretty big change from pure that actually makes this fork less pure.

I'll rather gate that with a variable.

mafredri:
@intelfx 

> [...] yet they can take >1 minute each on big repositories and cold VFS caches [...]

I haven't found this to be true. Could you give some examples? Pure uses `git rev-list --left-right --count HEAD...@'{u}'` and even on the linux tree this takes me barely any time at all.

> How exactly?

As in I see the name change from e.g. `master` -> `new-branch`. Don't get me wrong, it's quite quick, but still noticeable for me.

> Yes, the placeholders (`?w` and `?u`) are not particularly beautiful, but I wanted to highlight that the data is not yet ready rather than just show nothing and then let the `*` and arrows appear out of nowhere.

I can relate, it's not my meaning to criticize though, just providing feedback from my view. I've experimented with a loading indicator for pure as well (e.g. rotating dots) but didn't find anything I was happy with. Personally I think a single visual cue that everything isn't ready yet is the best (purest, heh?) way but YMMV.

> Could you please elaborate? [...]

I recorded [this effect](https://asciinema.org/a/45myroz25dd7of4dl4phoob7f) on asciinema, only the last prompt seems cached.

> I'll rather gate that with a variable.

You don't have to go that far, but if you want to, that's cool :).

intelfx:
@mafredri 

> I haven't found this to be true. Could you give some examples? Pure uses git rev-list --left-right --count HEAD...@'{u}' and even on the linux tree this takes me barely any time at all.

The below is done on a really fast machine with an SSD. If I ran this on a slower one with an HDD (have no access to that machine atm), this would take > 1 minute.

```
$ git describe --tags HEAD
v2.6.32.71
$ git describe --tags @{u}
v4.8-rc8-27-ge3b3656
$ sync; sudo sysctl vm.drop_caches=3; sync; sudo sysctl vm.drop_caches=3
vm.drop_caches = 3
vm.drop_caches = 3
$ time git rev-list --left-right --count HEAD...@{u}
4258    448998

real    0m14,351s
user    0m4,554s
sys     0m2,457s
```

Hint: to have `git rev-list --left-right` run slow, you need to have some divergence with the upstream :)

> As in I see the name change from e.g. master -> new-branch. Don't get me wrong, it's quite quick, but still noticeable for me.

I still don't understand. If the indicated branch name changes, then the last command indeed _did change_ the branch, correct? Then the indicated branch name will get updated as soon as `vcs_info` worker finishes. It is not guaranteed that it will finish first. For the meantime, I decided to print the old (stale) branch name to avoid even more visual stress due to changing line length too much.

mafredri:
Thanks for clarifying, I've never worked with this much divergence from upstream and did thus not know it was a problem (judging from our issues though, not a common problem 😛).

> I still don't understand.

I guess you're searching for the issue I'm referring to? There is none, it was just a remark as to the downsides of going fully async with `vcs_info`.

intelfx:
@mafredri 

> There is none, it was just a remark as to the downsides of going fully async with vcs_info.

Ah, yes. It has downsides. But `vcs_info` (which uses git-describe-ish things internally) is anything but fast, so I would never make that synchronous again.

mafredri:
@intelfx yup, I've looked at `vcs_info` as well and it does indeed do some wasteful operations. Many times I've considered just scrapping it, but I've decided to keep it around for simplicity (abstraction) / possibility of supporting other vcs than git. I may have to rethink this approach though.

intelfx:
@mafredri In particular, `vcs_info` does a variant of `git describe HEAD` instead of just reading branch name from `$GIT_DIR/HEAD` (and I like it this way). But this alone is enough to make it not constant-time.

mafredri:
Sorry for taking this long to merge @intelfx, forgot about it whilst we were having this interesting discussion ^_^.

PR #252, opened 2016-12-01
preprompt the last parent directory

Kikobeats:
What do you think?

screen shot 2016-12-01 at 22 01 55

Instead of 

screen shot 2016-12-01 at 22 02 26

moar minimalistic

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -137,7 +137,7 @@ prompt_pure_preprompt_render() {
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# construct preprompt, beginning with path
-	local preprompt="%F{blue}%~%f"
+	local preprompt="%F{blue}$(basename $PWD)%f"
 	# git info
 	preprompt+="%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f"
 	# git pull/push arrows


mafredri:
I don't think this is suitable for pure. This would hide important information and make it inconvenient to find out which directory your in. Would go against pures philosophy of staying out of the way.

sindresorhus:
What @mafredri said. Having the full path always visible was one of the main motivations of making Pure.

jiayihu:
I really need this feature for some projects, where frontend code is deep in the structure. Is there any workaround if you don't plan to support this option?

Kikobeats:
@sindresorhus what about control this behavior using a environment variable, like:

```bash
PURE_PREPROPMPT='%~'
PURE_PREPROPMPT='$(basename $PWD)'
```

mafredri:
@jiayihu, @Kikobeats changing this behavior and adding a configuration option are both non-starters. Care has been taken to make sure that Pure functions correctly even when the preprompt is multi-line.

My recommendation: use `setopt auto_name_dirs` and alias all your long paths as a variable.

Example `.zshrc`:
```zsh
setopt auto_name_dirs
GM=$GOPATH/src/github.com/mafredri

prompt pure
```

Now, here's what happens:

```console
~
❯ cd $GM

~GM
❯ pwd
/Users/mafredri/Golang/src/github.com/mafredri
```

For a bonus round, add `setopt cdable_vars` and you can simply `cd GM` without the `$`.

If you're still not satisfied, I'm afraid your only option is forking / hacking pure.

mafredri:
Added to wiki: [Create aliases for really long directory paths](https://github.com/sindresorhus/pure/wiki#create-aliases-for-really-long-directory-paths).

PR #253, opened 2016-12-04
Display virtualenv in prompt, if enabled

jmagnusson:


jmagnusson:
Whoops. Be right back with a new PR...

Issue #221, opened 2016-07-30
how add pyenv virtual environment display?


gatspy:
now ,i use pure theme and prezto, this theme how to support the work of pyenv?

mafredri:
Here's something I quickly hacked together which should work assuming you have initialized `pyenv` in your shell (`eval "$(pyenv init -)").

```shell
precmd() {
    if [[ -n $PYENV_SHELL ]]; then
        local version
        version=${(@)$(pyenv version)[1]}
        if [[ $version = system ]]; then
            PROMPT="%(?.%F{magenta}.%F{red})❯%f "
        else
            PROMPT="(pyenv $version) %(?.%F{magenta}.%F{red})❯%f "
        fi
    fi
}
```

1. We check if `pyenv` has been initialized (the `$PYENV_SHELL` parameter has been set)
2. We get active the pyenv version by running `pyenv version`. With `(@)` we are turning the output into an array and with `[1]` we fetch the first element.
3. If the active python is the `system` version, we just use the default prompt. If it's not system, we prefix the prompt with it.

---

I think this is perhaps something we even could support in `pure`, what are your thoughts on the matter @sindresorhus? I'm not entirely sure on how to present the information, this is what it looks like now, but I think we can do better if we decide to support it:

![](https://www.dropbox.com/s/qi23omy94nmp9u9/Screenshot%202016-07-30%2018.42.23.png?dl=1)

gatspy:
This is indeed a good idea, thanks @mafredri

sindresorhus:
Sure, looks simple enough.

z0rc:
The proposed approach is way too heavy (two forks). You can just check for presence of `PYENV_VERSION` shell var and output it if it's set.

But given that there are multiple tools to manage virtualenvs, I'd propose to check for `VIRTUAL_ENV` and output it as `${VIRTUAL_ENV##*/}` to show only last folder from pull path stored in that variable.

mafredri:
Sorry for taking so long to get back to you on this @z0rc, I'm aware the reference implementation here is not very efficient, and would be implemented differently in pure.

I'm not entirely confident what you're suggesting would work, though. As far as I can tell, the `PYENV_VERSION` env var does not get set by itself, example:

```
~
❯ eval "$(pyenv init -)"

~
❯ python --version
Python 2.7.12

~
❯ typeset -p PYENV_VERSION
typeset: no such variable: PYENV_VERSION

~
❯ cd derp

~/derp
[...]

PR #254, opened 2016-12-04
Display virtualenv in prompt if enabled

jmagnusson:


mafredri:
I'm not entirely sold on the placement / color of the virtualenv info. We've been considering virtualenv support for a while now, e.g. in #221. Unfortunately I haven't had too much time to dedicate to this lately.

Since we've previously only kept git information in the pre-prompt, there hasn't been a warrant to explain what information we are showing. If virtualenv was to be added, I feel it's not quite sufficient to just add a new color in the mix. For this reason I've favored the idea of prefixing the prompt character with the virtualenv. This is how virtualenvs have traditionally done it, and visually I think it's the best representation of the fact that the prompt is now in a different environment.

For reference, here's a visual for why I don't think the placement / color really works in this PR:
screenshot 2016-12-05 17 54 23

jmagnusson:
> We've been considering virtualenv support for a while now, e.g. in #221.

#221 is for `pyenv`, not plain `virtualenv`. Far from everyone is using `pyenv`.

> and visually I think it's the best representation of the fact that the prompt is now in a different environment.

Using that line of argumentation would mean that anything that prepends anything to `$PATH` should be put on the prompt row.

> For reference, here's a visual for why I don't think the placement / color really works in this PR

Execution time displayed in yellow just next to the virtualenv name is usually not a problem I'd bet as one typically would activate them while inside a git repository, which name would be put between venv name and execution time, with a different color. I've been using this PR for two days now and have yet to be distracted by it.

Personally what I like most about the `pure` prompt is just that - it's simplistic and pure. Adding the venv name to the prompt row would kind of feel like a departure from what makes `pure` so great.

mafredri:
> #221 is for `pyenv`, not plain `virtualenv`. Far from everyone is using `pyenv`.

The issue is titled `pyenv` but we've discussed both in there. I see them as related issues. Just supporting `$VIRTUALENV` would also support a large chunk of `pyenv` users.

> Using that line of argumentation would mean that anything that prepends anything to `$PATH` should be put on the prompt row.

I disagree, a virtualenv is not about modifying the path (what technically happens in the background is irrelevant). Virtualenv is a tool to create isolated Python environments, as such it hijacks your environment and replaces your regular software with it's own. I think it's fair to think of it as a different environment. Ultimately I wish not to argue about how we see virtualenvs, but I'd like to point out that I don't see this as a good argument for the dismissal of "virtualenv behind prompt".

> Execution time displayed in yellow just next to the virtualenv name is usually not a problem I'd bet as one typically would activate them while inside a git repository, which name would be put between venv name and execution time, with a different color. I've been using this PR for two days now and have yet to be distracted by it.

I'm not really satisfied with something not *usually* being a problem. I'm strongly of the opinion that a yellow virtualenv and a yellow execution time has a chance of being confusing and thus not a good fit with pure. Furthermore, the yellow color often stands out much, much more, especially with the git part being a darkened gray. I say often because it is, of course, dependent on the terminal color configuration, but for me the yellow is quite distracting and makes the git information almost invisible.

Some thoughts I think need to be considered:
* Should the virtualenv really stand out more than the git information (yellow vs dark gray)?
* What would be an optimal color?
* What's the best position for it?
    * Preprompt: before path, where it's now or after git?
    * Prompt: if obtrusive, can it be made not-so?
* Is the name of the virtualenv the optimal thing to display? How about just indicating that a virtualenv is active with a character?
* Maybe it could be displayed differently, e.g. `(virtualenv)` using the same color as git?

> Personally what I like most about the `pure` prompt is just that - it's simplistic and pure. Adding the venv name to the prompt row would kind of feel like a departure from what makes `pure` so great.

Could you please elaborate, why do you think so? What aspect of it is less simplistic/pure?

---

Finally, two screenshots to illustrate some ideas I talked about above:

screenshot 2016-12-07 20 42 22
screenshot 2016-12-07 20 48 13

seemethere:
👍 Or at least have an option to enable it

andrroy:
@mafredri What is the status of this one? Is there something I can do to help get this merged?

mafredri:
@andrroy currently the color choice, and possibly the placement are blocking this PR, also, I'm still of the opinion that this would be better on the prompt line, in front of the prompt (e.g. `venv ❯ `). There are lots of undiscussed points in my post above.

andrroy:
For what its worth, I've added my opinions bellow:

> Should the virtualenv really stand out more than the git information (yellow vs dark gray)?

I don't think that the virtualenv should draw more attention than the git information, so i would prefer it to be gray

> What would be an optimal color?

As mentioned above, I would prefer the same gray color as the git information

> What's the best position for it?

I'm open to both propositions. The line above the prompt can become a bit long if you're deep down in a directory, and working on a branch with a long name. So i think I would prefer having it on the same line as the prompt. 

I can see your point @jmagnusson, but in my case I only source environments while working on something specific. As soon as I'm done, I deactivate the env. 

> Is the name of the virtualenv the optimal thing to display? How about just indicating that a virtualenv is active with a character?

It's important to know what environment you are working on, especially for users that use virtualenvwrapper.  It should at least be an option to enable it.

mortenvn:
I sat down trying to figure out the best way to solve this. In the end I came to the same conclusion as @mafredri and @andrroy: It would be best to just display the virtualenv name on the prompt line (in grey).

just-name

---

There are several reasons why I think adding it to the prompt (and not the preprompt) is the better solution: 

**Easier to "parse" when the preprompt is long**
Let's consider how things would look when the preprompt gets long:
original

While the original proposal looks pretty, I feel that the preprompt becomes a bit cluttered and the venv_name easy to miss. By contrast, when placed in the prompt it remains visible and understandable no matter how long the preprompt is.

**Avoids confusion**
> I'm strongly of the opinion that a yellow virtualenv and a yellow execution time has a chance of being confusing and thus not a good fit with pure. 

I 100% agree with this. And I'm not convinced that just changing the color would be a good solution either as the preprompt becomes slightly fruitcake-ish:
original-alt-color

**Similar to default look**
The default implementation places the venv_name at the beginning of the prompt. I think placing it in a similar location would be more in line with the users expectations.
venv-activated2

**Virtualenv changes commands behaviour**
The way I see it, using virtualenv might make your commands behave differently than they otherwise would. I think it's useful to be reminded of this just before typing your commands.


---
**Alternative solutions with just an icon**
I tried pursuing the idea of using just an icon. One solution I tried was adding an extra prompt symbol. However, I don't think this solution is obvious enough.
just-icon

I looked at a lot (A LOT) of different unicode icons, but I couldn't really find anything that would make things more obvious.
screen shot 2017-05-06 at 01 08 44

Even if we would find the **perfect** icon, I still think people would miss seeing the actual environment name.

**P.S.** Really loving the Pure theme. Thanks for being awesome! ❤️

mafredri:
Thanks for your input @andrroy and @mortenvn! It's really helpful to hear thoughts and feedback from people who actively use virtualenvs. Also, I really appreciate the effort you put in @mortenvn, it really helps narrow down some of these options (PS. love the reference 😉)!

> It would be best to just display the virtualenv name on the prompt line (in grey).

I'm leaning towards this as well. Since you've done a bunch of testing, would you like to submit this as a new PR (bonus points if you can consider the impact when combined with #323 as well!)? No pressure though 😄.

Feels like we're finally close to putting a lid on this virtualenv saga, thanks guys!

mortenvn:
Fantastic! Glad to hear the feedback of use 😃

> Since you've done a bunch of testing, would you like to submit this as a new PR?

Sure! I'd love to. Just give me a couple of days.


> (bonus points if you can consider the impact when combined with #323 as well!)

I have a few ideas. Dunno if they are any good yet. 😛 
I'll do some testing and get back to you.

mafredri:
Closing this now that #325 has been merged.

PR #256, opened 2016-12-05
Add support for custom prompt arrow color

petermbenjamin:
### Feature
Support custom prompt arrow color when commands complete or fail.

This can be controlled via the following options:
`PURE_PROMPT_COLOR_NORMAL` and `PURE_PROMPT_COLOR_FAIL`

### Reasoning
- There is not enough contrast in the colors between the default color for successful commands (`magenta`) and the default color for failed commands (`red`). 
- In fact, for color-blind folks, `red` and `magenta` appear as the same color.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -364,7 +364,7 @@ prompt_pure_setup() {
 	[[ $UID -eq 0 ]] && prompt_pure_username=' %F{white}%n%f%F{242}@%m%f'
 
 	# prompt turns red if the previous command didn't exit with 0
-	PROMPT="%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f "
+	PROMPT="%(?.%F{${PURE_PROMPT_COLOR_NORMAL:-magenta}}.%F{${PURE_PROMPT_COLOR_FAIL:-red}})${PURE_PROMPT_SYMBOL:-❯}%f "
 }
 
 prompt_pure_setup "$@"
--- a/readme.md
+++ b/readme.md
@@ -109,6 +109,14 @@ Defines the git down arrow symbol. The default value is `⇣`.
 
 Defines the git up arrow symbol. The default value is `⇡`.
 
+### `PURE_PROMPT_COLOR_FAIL`
+
+Defines the prompt arrow color when commands fail. The default is `red`.
+
+### `PURE_PROMPT_COLOR_NORMAL`
+
+Defines the prompt arrow color when commands complete successfully. The default is `magenta`.
+
 ## Example
 
 ```sh


seemethere:
:+1:

sindresorhus:
What would you set those options to? Just curious.

seemethere:
For me it'd be green for success and red for fail. It goes better with the color scheme I am using right now.

pmbenjamin:
@Sindresorhus
Green = success 
Red = fail

Rather than
Magenta = success
Red = fail

Because magenta/red combo don't have enough color contrast.

mafredri:
We get a few requests for modifying the colors (and symbols, etc) of Pure through options every now and a then. The reason we do not accept them is that we don't want to clutter pure with options. Ideally Pure would have even less options than it does currently. 

It's actually possible to override the `PROMT` var after pure has been initialized. There are of course no guarantee that this will always be the case (in case of new features requiring modifications to the promt). Not that this is any better than an option (relying on hidden knowledge), pus its possible without forking the theme. 

In my opinion, merging this PR would mean that we need to embrace full custom colorization of the prompt. This in turn means new options or an alternate method of defining the colors. 

I also don't mean to be disrespecting of any level of color blindness, but if I had a problem telling the colors apart, I would change them in my terminal so that they are more so (no reason to keep magenta as magenta). 

All this being said, I have no strong preference on the default promt color(s). Logically green/red makes most sense, but I took a liking to magenta after using it with Pure (as @sindresorhus intended 😝).

PS. Entertaining that this PR would be merged, I think variable names should not be `NORMAL/FAIL`, instead `STATUS_OK/STATUS_ERROR` (or even status zero/nonzero).

pmbenjamin:
I appreciate your feedback @mafredri. 
I will respect the consensus of the community. 

My goal was to simply contribute back a minor change I had made, in case others wanted to enjoy the same benefits.

And since the prompt character itself (the arrow) was open to customization via variables, I took the same approach to customize the colors of said character.

peterkc:
The variable names of `STATUS_OK/STATUS_ERROR` are not ambiguous–clearly states that they pertain to the status.

pmbenjamin:
If the maintainer thinks this PR is a good contribution, I can update the environment variables to `STATUS_OK`/`STATUS_ERROR`.

craftyguy:
I have a VERY difficult time differentiating between the magenta and red colors, especially when they are represented by a single character. If this PR is not accepted, then please consider using a different color other than magenta for "return value OK".

dcastil:
Would like to see this being implemented, too! I also have a hard time to distinguish between magenta and red and I believe form should follow function here!

petermbenjamin:
I fixed the merge conflicts on this PR.

craftyguy:
For what it's worth, I've been running with this patch for over a year now, on several systems, and have had  0 problems. Being able to set the color to something that is easier to distinguish from the red arrow prompt is great. Thank you @petermbenjamin, I hope this gets merged so that others can benefit!

sindresorhus:
Closing this in favor of https://github.com/sindresorhus/pure/pull/472.

PR #259, opened 2016-12-12
Readme improvements

mafredri:
Two changes here:

1. We will now require zsh 5.2+ from this point onwards (reasoning in https://github.com/sindresorhus/pure/commit/d3367ec20bfcbaa230ec8730c38d18a3d3de2bf5)
2. Oh-my-zsh
    * Instructions changed to prevent collision with bundled themes
    * Provide alternate method to avoid integration entirely.

@sindresorhus does this look ok, or can you think of some improvements still?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,7 +24,7 @@ Most prompts are cluttered, ugly and slow. I wanted something visually pleasing 
 
 ## Install
 
-Can be installed with `npm` or manually. Requires git 2.0.0+ and ZSH 5.0.0+.
+Can be installed with `npm` or manually. Requires git 2.0.0+ and zsh 5.2+. Older versions of zsh are known to work, but they are **not** recommended.
 
 ### npm
 
--- a/readme.md
+++ b/readme.md
@@ -24,7 +24,7 @@ Most prompts are cluttered, ugly and slow. I wanted something visually pleasing 
 
 ## Install
 
-Can be installed with `npm` or manually. Requires Git 2.0.0+ and ZSH 5.2+. Older versions of zsh are known to work, but they are **not** recommended.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -24,7 +24,7 @@ Most prompts are cluttered, ugly and slow. I wanted something visually pleasing
 
 ## Install
 
-Can be installed with `npm` or manually. Requires git 2.0.0+ and ZSH 5.0.0+.
+Can be installed with `npm` or manually. Requires git 2.0.0+ and zsh 5.2+. Older versions of zsh are known to work, but they are **not** recommended.
```
sindresorhus: zsh => ZSH
git => Git

mafredri: Sure, I was following the `zsh` capitalization used in the zsh docs, but I don't mind changing this.

readme.md
```diff
@@ -137,8 +137,14 @@ To have commands colorized as seen in the screenshot, install [zsh-syntax-highli
 
 ### [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh)
 
-1. Symlink (or copy) `pure.zsh` to `~/.oh-my-zsh/custom/themes/pure.zsh-theme`.
-2. Set `ZSH_THEME="pure"` in your `.zshrc` file.
+1. Symlink (or copy) `pure.zsh` to `~/.oh-my-zsh/custom/sindresorhus-pure.zsh-theme`.
+2. Symlink (or copy) `async.zsh` to `~/.oh-my-zsh/custom/async.zsh`.
+3. Set `ZSH_THEME="sindresorhus-pure"` in your `.zshrc` file.
```
mafredri: Oh, didn't know you had started this. Good effort!

Looks like it isn't PR'd yet, but if it gets merged then this definitely is not necessary. I'll change it back to just `pure` for now. Let's revisit if the oh-my-zsh version never gets renamed.

sindresorhus:
Looks good :)

PR #255, opened 2016-12-04
Add zplug install instructions

gillesdemey:

PR #261, opened 2016-12-13
Add zplug install instructions

seemethere:
Adds the install instructions for zplug.

Separates the two for installing zsh-async and pure which #255 does not do

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -166,11 +166,11 @@ $ antibody bundle sindresorhus/pure
 
 ### [zplug](https://github.com/zplug/zplug)
 
-Update your zshrc with the following lines
+Update your `.zshrc` file with the following lines (order matters):
 
-```shell
-zplug "mafredri/zsh-async", from:github
-zplug "sindresorhus/pure", use:pure.zsh, from:github, as:theme
+```console
+$ zplug mafredri/zsh-async, from:github, defer:0  # Load this first
+$ zplug sindresorhus/pure, use:pure.zsh, from:github, as:theme
 ```
 
 ## FAQ
--- a/readme.md
+++ b/readme.md
@@ -164,6 +164,15 @@ $ antibody bundle mafredri/zsh-async
 $ antibody bundle sindresorhus/pure
 ```
 
+### [zplug](https://github.com/zplug/zplug)
+
+Update your zshrc with the following lines
+
+```shell
+zplug "mafredri/zsh-async", from:github
+zplug "sindresorhus/pure", use:pure.zsh, from:github, as:theme
+```
+
 ## FAQ
 
 ### My preprompt is missing when I clear the screen with Ctrl+L


sindresorhus:
What's the benefit of separating them? I prefer #255 for its simplicity.

// @gillesdemey

seemethere:
Also zsh-async could be used for something else. Or maybe zsh-async is updated before it's updated here. It's to handle them as separate dependencies which, from a user perspective, is better for both understanding and management of updates.

sindresorhus:
What if `zsh-async` releases a version that doesn't work with the current version of Pure? There's an obvious benefit of having them bundled together, the `zsh-async` version is locked and tested together with Pure.

seemethere:
But it crushes any version of zsh-async that another plugin may need. Pure may not be the only plugin that uses zsh-async.

sindresorhus:
Yes, that's the problem with global dependencies, but at least Pure will work this way. Your way, it might break all plugins using `zsh-async`.

@mafredri Any way we could make `zsh-async` async a "local" dependency for Pure? So at least it would conflict with other consumers of `zsh-async`.

mafredri:
@sindresorhus honestly, not easily. All function names in zsh-async would need to be prefixed in Pure so as to not conflict with the original. This would probably mean introducing a build-pipeline for Pure (e.g. `src/{async.zsh,pure.zsh}` -> `pure.zsh`), otherwise updating zsh-async could become cumbersome.

My thought has always been to provide zsh-async in the repo for convenience, and Pure, in turn, respects the users choice by using `autoload` to load `async` from the `$fpath`. Unless I'm mistaken, #255 sources the shell scripts and would thus overwrite any `async` that the user might be hoping to load into his shell.

I don't see this as a big problem, tough, as I always test any updates to zsh-async with Pure for a while before I do a new release (even pushing to master). Also, I have no plans to change the zsh-async API, so any updates are likely to be bug / compatibility fixes.

sindresorhus:
@mafredri Ok. Which way do you prefer, this PR or #255?

mafredri:
@sindresorhus for now I would prefer this PR since it's similar to the current instructions for `antigen` and `antibody` and more respecting of autoload behavior.

Sidenote: #255 should probably be `use:"{async,pure}.zsh"` instead of `use:"*.zsh"` since we also have the `pure.plugin.zsh` symlink in the repo.

readme.md
```diff
@@ -164,6 +164,15 @@ $ antibody bundle mafredri/zsh-async
 $ antibody bundle sindresorhus/pure
 ```
 
+### [zplug](https://github.com/zplug/zplug)
+
+Update your zshrc with the following lines
```
sindresorhus: Should be:

```
Update your `.zshrc` file with the following lines (order matters):
```

readme.md
```diff
@@ -164,6 +164,15 @@ $ antibody bundle mafredri/zsh-async
...
+
+```shell
```
sindresorhus: console

readme.md
```diff
@@ -164,6 +164,15 @@ $ antibody bundle mafredri/zsh-async
...
+zplug "mafredri/zsh-async", from:github
+zplug "sindresorhus/pure", use:pure.zsh, from:github, as:theme
```
sindresorhus: Prefix the lines with `$`

sindresorhus: Unquote these if possible

readme.md
```diff
@@ -164,6 +164,15 @@ $ antibody bundle mafredri/zsh-async
 $ antibody bundle sindresorhus/pure
 ```
 
+### [zplug](https://github.com/zplug/zplug)
+
+Update your `.zshrc` file with the following lines (order matters):
+
+```console
+$ zplug mafredri/zsh-async, from:github, defer:0  # Load this first
+$ zplug sindresorhus/pure, use:pure.zsh, from:github, as:theme
```
mafredri: @sindresorhus not really relevant for this PR, but in my opinion the `console` and `$` don't really make sense here since they represent lines in a file (`.zshrc`).

Thoughts? If you agree I can submit a PR to change this for all the examples where it doesn't make sense.

sindresorhus: 👍

PR #271, opened 2017-01-16
Update zsh-async to 1.5.0

mafredri:
This commit updates `zsh-async` to the latest versions.

Short list of changes:

* 1.4.0: Many issues and bugs were fixed (caught by new test suite)
* 1.5.0: Better support for legacy zsh (5.0.2, 5.0.8, 5.1.1), no more reliance on kill-signals to notify of completed tasks

See zsh-async [releases](https://github.com/mafredri/zsh-async/releases) for more details.

PR #272, opened 2017-01-16
Check for git arrows asynchronously

mafredri:
Previously `command git rev-list --left-right --count HEAD...@'{u}'` was run in a subshell, and upon `git fetch` completion, run once again.

This is now done as an async job (`prompt_pure_async_git_arrows`) and is also called within `prompt_pure_async_git_fetch` to return the arrow status immediately after `git fetch` has completed, this saves us from waiting until the git fetch is done to only then execute a new async
job to check the arrow status.

Changes in git arrow / git dirty status are also handled better and `prompt_pure_preprompt_render` is not called unless they actually changed.

Some things to note:

* I've started using `setopt localoptions noshwordsplit` in some functions to allow using variable names without quoting (ideally we should update other functions as well, in a separate PR)
* This PR depends on #271 and allows for some simplifications, e.g. `builtin cd -q "$*"` -> `builtin cd -q $1`, due to improvements in zsh-async
* `prompt_pure_check_git_arrows` now assigns the result to `$REPLY` instead of taking a variable name to store the result in, I kind of prefer this (many built-in zsh functions also assign to `$REPLY`) but am open to changing it to match our other function signatures.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -61,29 +61,6 @@ prompt_pure_clear_screen() {
 	prompt_pure_preprompt_render precmd
 }
 
-prompt_pure_check_git_arrows() {
-	# reset git arrows
-	prompt_pure_git_arrows=
-
-	# check if there is an upstream configured for this branch
-	command git rev-parse --abbrev-ref @'{u}' &>/dev/null || return
-
-	local arrow_status
-	# check git left and right arrow_status
-	arrow_status="$(command git rev-list --left-right --count HEAD...@'{u}' 2>/dev/null)"
-	# exit if the command failed
-	(( !$? )) || return
-
-	# left and right are tab-separated, split on tab and store as array
-	arrow_status=(${(ps:\t:)arrow_status})
-	local arrows left=${arrow_status[1]} right=${arrow_status[2]}
-
-	(( ${right:-0} > 0 )) && arrows+="${PURE_GIT_DOWN_ARROW:-⇣}"
-	(( ${left:-0} > 0 )) && arrows+="${PURE_GIT_UP_ARROW:-⇡}"
-
-	[[ -n $arrows ]] && prompt_pure_git_arrows=" ${arrows}"
-}
-
 prompt_pure_set_title() {
 	# emacs terminal does not support settings the title
 	(( ${+EMACS} )) && return
@@ -216,9 +193,6 @@ prompt_pure_precmd() {
 	# with the initial preprompt rendering
 	prompt_pure_cmd_timestamp=
 
-	# check for git arrows
-	prompt_pure_check_git_arrows
-
 	# shows the full path in the title
 	prompt_pure_set_title 'expand-prompt' '%~'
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Been using this for a couple of days now and seems to work well :)

> I've started using setopt localoptions noshwordsplit in some functions to allow using variable names without quoting (ideally we should update other functions as well, in a separate PR)

:+1:

> prompt_pure_check_git_arrows now assigns the result to $REPLY instead of taking a variable name to store the result in, I kind of prefer this (many built-in zsh functions also assign to $REPLY)

Just curious, how does `$REPLY` work? I've never used that pattern before.

mafredri:
> Been using this for a couple of days now and seems to work well :)

Good to hear! :)

> Just curious, how does `$REPLY` work? I've never used that pattern before.

When working with Zsh builtins and/or expansion, sometimes one of two variables are used to return the value, `$REPLY` (for values/strings) and `$reply` for arrays. One such example is the `read` command:

```console
❯ print pure | read
❯ print $REPLY
pure
```

You can provide a variable `name` to read, though, in which case it would be stored in that variable instead of `$REPLY`. There are many [examples](https://github.com/zsh-users/zsh/blob/a4b8ee13be494c8732c0348857befe87951a058b/Functions/Chpwd/zsh_directory_name_cdr#L4-L13) [in](https://github.com/zsh-users/zsh/blob/a4b8ee13be494c8732c0348857befe87951a058b/Functions/Chpwd/chpwd_recent_add#L20-L24) [the](https://github.com/zsh-users/zsh/blob/a4b8ee13be494c8732c0348857befe87951a058b/Completion/Base/Utility/_regex_words#L23-L49) [zsh](https://github.com/zsh-users/zsh/blob/a4b8ee13be494c8732c0348857befe87951a058b/Functions/Zle/split-shell-arguments#L9) [repository](https://github.com/zsh-users/zsh/blob/a4b8ee13be494c8732c0348857befe87951a058b/Functions/Zle/read-from-minibuffer#L50).

The idea here is that the callee (in our case `prompt_pure_check_git_arrows`) sets `$REPLY` (an agreed upon naming/pattern) on the global scope as a way of returning the value it produces. However, we don't really want to leak the value to the global scope since this would lead to the user suddenly having a value in `$REPLY` that originated from some unknown place. 

We can avoid leaking this value by defining `local REPLY` in the caller because the scope of the callee is always that of the caller. In this situation the global `$REPLY` for `prompt_pure_check_git_arrows` is actually the local `$REPLY` defined in `prompt_pure_async_callback`.

sindresorhus:
@mafredri Thanks for the explanation. You should really do a blog post about this ;) I find it very hard to Google ZSH concepts like this.

mafredri:
No problem, hope it was palatable :D. I guess I need to set up a blog 😝.

sindresorhus:
@mafredri Would recommend just going with medium.com. Spend your time writing, not setting up a blog ;)

sindresorhus:
I would be happy to review by the way.

mafredri:
Thanks, I just might take you up on that offer should I manage to put something together ^_^!

Issue #143, opened 2015-07-13
Multiline ZLE buffer overwritten by preprompt update


zmwangx:
A screencast should suffice to demonstrate this problem:

[![asciicast](https://asciinema.org/a/645061r08iwz94hdxxtnfkmpz.png)](https://asciinema.org/a/645061r08iwz94hdxxtnfkmpz)

The problem is that the [`lines` variable in `prompt_pure_preprompt_render`](https://github.com/sindresorhus/pure/blob/5f5b990e2b1739ba359f08a33727c8bf107bf4f8/pure.zsh#L108) doesn't take into account current ZLE buffer lines (it only records the number of lines spanned by the preprompt), so if the buffer spans multiple lines, then the preprompt is updated in the wrong place.
## Proposed fix

A naive fix: ZLE has a read-only parameter `BUFFERLINES` available to ZLE widgets. We can thus register a ZLE widget, say `prompt_pure_record_buffer_lines`, that records `BUFFERLINES` in a global variable, say `prompt_pure_buffer_lines`. Then we can simply do `(( lines += prompt_pure_buffer_lines - 1 ))` to account for the buffer.

However, there are two other technical challenges:
1. The naive fix doesn't capture the case where the cursor is not at the end of the buffer. But this is easy to deal with. We can determine the absolute position of the cursor (from the beginning of the line) by summing the length of `PS1/PS2` (depending on `CONTEXT`) and the `CURSOR` parameter. (We already have a function `prompt_pure_string_length` to calculated the prompt-expanded length of a string.) Then we can determine the line where cursor is on by dividing by `COLUMNS`.
2. The naive fix only works when the ZLE `CONTEXT` is `start`; if the `CONTEXT` is, say, `cont` (continuation of a command line), then the previous lines would have already been flushed from the ZLE buffer, so `BUFFERLINES` doesn't count those lines. As a concrete example, consider
   
   ```
   > echo "hello
   dquote> world"_
   ```
   
   where `_` represents the cursor position. At that point, `BUFFERLINES` is 1, but we should count both lines in order to go back to the preprompt.
   
   To deal with this scenario, the solution I came up with is to use an accumulator, say `prompt_pure_accumulated_buffer_lines`, which is cleared precmd and incremented by `BUFFERLINES` right before each `accept-line`. This would require replacing the default `accept-line` with a custom widget.

[...]

PR #273, opened 2017-01-17
Perform all git checks (vcs_info) asynchronously

mafredri:
This is a quick experiment to perform `vcs_info` asynchronously. For anyone trying this out, would appreciate some feedback on how it feels / behaves / problems / etc.

Here we try to detect if the git toplevel has changed so that we can immediately clear the git branch / dirty / arrow status. The problem is, `git rev-parse --show-toplevel` (as used by `vcs_info`) might differ from `$PWD` (e.g. because of symlinks).

To mitigate the above problem to some extent we store the current `$PWD` when entering a git project, this `$PWD` is then used in future comparisons to see if the new `$PWD` is at least a partial match of `$PWD`, if not everything is cleared immediately.

To illustrate what happens:

1. `cd ~/pure`
2. `prompt_pure_async_vcs_info` detects a new git top level `/home/user/pure`, `prompt_pure_vcs_info[pwd]=/home/user/pure`
3. cd `arch`
4. `[[ /home/user/pure/arch =~ ^/home/user/pure ]]` is true, meaning we did not likely change git toplevel, keep the branch / arrow / dirty status

And in the case of first entering a sub folder:

1. `cd ~/pure/arch`
2. `prompt_pure_async_vcs_info` detects a new git top level `/home/user/pure`, `prompt_pure_vcs_info[pwd]=/home/user/pure/arch`
3. cd `..`
4. `[[ /home/user/pure =~ ^/home/user/pure/arch ]]` is not true, so git info is cleared
5. `prompt_pure_async_vcs_info` returns and restores the git info, this time setting `prompt_pure_vcs_info[pwd]=/home/user/pure`

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -379,7 +379,8 @@ prompt_pure_async_callback() {
 			local -A info
 			typeset -gA prompt_pure_vcs_info
 
-			info=("${(Q@)=output}")
+			# parse output (z) and unquote as array (Q@)
+			info=("${(Q@)${(z)output}}")
 			local -H MATCH
 			# check if git toplevel has changed
 			if [[ $info[top] = $prompt_pure_vcs_info[top] ]]; then
--- a/pure.zsh
+++ b/pure.zsh
@@ -200,7 +200,7 @@ prompt_pure_preprompt_render() {
 		else
 			# When lines equal, we can try to perform a partial prompt update.
 			integer pos
-			for (( pos = 1; pos < $#preprompt_parts; pos++ )); do
+			for (( pos = 1; pos <= $#preprompt_parts; pos++ )); do
 				patch_pos=$pos
 				if [[ $preprompt_parts[$pos] != $prompt_pure_last_preprompt[$pos] ]]; then
 					break  # We found where the preprompts differ.
--- a/pure.zsh
+++ b/pure.zsh
@@ -106,151 +106,64 @@ prompt_pure_string_length_to_var() {
 }
 
 prompt_pure_preprompt_render() {
-	# store the current prompt_subst setting so that it can be restored later
-	local prompt_subst_status=$options[prompt_subst]
+	setopt localoptions noshwordsplit
 
-	# make sure prompt_subst is unset to prevent parameter expansion in preprompt
-	setopt local_options no_prompt_subst no_sh_word_split
+	# Check that no command is currently running, the preprompt will otherwise
+	# be rendered in the wrong place.
+	[[ -n ${prompt_pure_cmd_timestamp+x} ]] && [[ $1 != precmd ]] && return
 
-	# check that no command is currently running, the preprompt will otherwise be rendered in the wrong place
-	[[ -n ${prompt_pure_cmd_timestamp+x} && "$1" != "precmd" ]] && return
-
-	# set color for git branch/dirty status, change color if dirty checking has been delayed
+	# Set color for git branch/dirty status, change color if dirty checking has
+	# been delayed.
 	local git_color=242
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# Initialize the preprompt array.
 	local -a preprompt_parts
 
-	# Expand the working directory for later comparison.
-	local expanded_pwd="%~"
-	expanded_pwd=${(%)expanded_pwd}
+	# Set the path.
+	preprompt_parts+=("%F{blue}%~%f")
 
-	preprompt_parts+=("%F{blue}$expanded_pwd%f")
-
-	# git info
+	# Add git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
 	if [[ -n $prompt_pure_vcs_info[branch] ]]; then
-		preprompt_parts+=(" %F{$git_color}${prompt_pure_vcs_info[branch]}%f")
-		preprompt_parts+=("%F{$git_color}${prompt_pure_git_dirty}%f")
+		preprompt_parts+=("%F{$git_color}\${prompt_pure_vcs_info[branch]}\${prompt_pure_git_dirty}%f")
 	fi
-	# git pull/push arrows
+	# Git pull/push arrows.
 	if [[ -n $prompt_pure_git_arrows ]]; then
-		preprompt_parts+=(" %F{cyan}${prompt_pure_git_arrows}%f")
+		preprompt_parts+=("%F{cyan}${prompt_pure_git_arrows}%f")
 	fi
 
-	# username and machine if applicable
+	# Username and machine, if applicable.
 	[[ -n $prompt_pure_username ]] && preprompt_parts+=($prompt_pure_username)
-	# execution time
+	# Execution time.
 	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=("%F{yellow}${prompt_pure_cmd_exec_time}%f")
 
-	# Join the preprompt array on empty string.
-	local preprompt=${(j..)preprompt_parts}
+	local -ah ps1
 
-	# if executing through precmd, do not perform fancy terminal editing
-	if [[ "$1" == "precmd" ]]; then
-		print -P "\n${preprompt}"
-	else
-		local prev_preprompt=${(j..)prompt_pure_last_preprompt}
-		if [[ $prev_preprompt == $preprompt ]]; then
-			# Avoid redrawing when there are no changes.
-			return
-		fi
+	# Construct the new prompt, containing preprompt.
+	PROMPT=${PROMPT//$prompt_newline/$'\n'}
+	ps1=(${(f)PROMPT})  # Split on newline.
+	ps1=(
+		$prompt_newline           # Initial newline, for spaciousness.
+		${(j. .)preprompt_parts}  # Join parts, space separated.
+		$prompt_newline           # Separate preprompt and prompt.
+		$ps1[-1]                  # Keep last part of the prompt.
+	)
 
-		# We need the length/lines of the preprompt to
-		# figure out what update method we should use.
-		integer preprompt_length lines
-		prompt_pure_string_length_to_var "$preprompt" preprompt_length
-		lines=$(( ((preprompt_length - 1) / COLUMNS) + 1 ))
+	PROMPT="${(j..)ps1}"
 
-		# We use the length/lines of the previous preprompt to
-		# figure out the best update method for the new one.
-		integer last_preprompt_length last_lines
-		prompt_pure_string_length_to_var "$prev_preprompt" last_preprompt_length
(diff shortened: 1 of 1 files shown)


sindresorhus:
I've been running this for a week now and haven't noticed any issues, but can't say I've noticed any speed improvement either, although my computer is pretty fast.

sindresorhus:
@pmbenjamin @seemethere @peterkc @jmagnusson Could you test this out?

mafredri:
Thanks for giving it a good spin! I'd say the biggest improvement in prompt responsiveness can see seen when issuing consecutive commands or when the system is otherwise under heavy load.

I've noticed one issue, when I rebased a branch without pushing the changes, rebranched it (checkout -b) so that there was no origin, the git arrows were not reset. I'll look to fix this soon.

Some other cases where I think this could provide better responsiveness is on spinning disks, and partitions mounted over the network (esp. sshfs) although I don't have any setups like this.

seemethere:
I'll give it a go @sindresorhus

sindresorhus:
@mafredri I discovered a bug. It fails when you try to enter a git repo where the directory name contains spaces:

```
~/Downloads
❯ cd 'foo bar'

~/Downloads/foo bar
❯
prompt_pure_async_callback:9: bad set of key/value pairs for associative array
❯
```

mafredri:
Good catch @sindresorhus! Should be fixed in d8de8f6.

seemethere:
Can't really say I've had any issues with this branch so far.

mafredri:
Thanks for testing @seemethere! Did you notice that it behaves differently in any way? Does it feel more responsive / sluggish?

The biggest difference here is that the branch name is updated after the prompt is rendered, meaning e.g. when you switch from master -> development, the next prompt will include master, which then almost immediately gets overwritten with a new prompt containing development. I imagine some people will notice this more than others.

I have an idea for improving it though, we could rewrite the algorithm in `prompt_pure_preprompt_render` to allow us to patch the prompt instead of a full update like we currently do. I believe this should speed up the branch name update but it's not a given, of course.

PS. I pushed a fix for the git arrows remaining after switching to a branch with no upstream.

mafredri:
Ok, so I just pushed a commit for performing partial prompt updates when possible. To me it feels slightly snappier, can you notice a difference (e.g. when switching branches)?

It could be further optimized by keeping track of the cursor position using zle widgets. This would allow us to optimize our use of ansi-escape codes to move the smallest amount of distance necessary to perform the update. Keeping track of the position would be a precursor for fixing #143 as well.

Not to worry 26f2266 is pretty big, but a large chunk of that is updated comments. The logic change is still pretty small.

mafredri:
I just amended my previous commit with 42c1ddb. Now the partial preprompt update is actually be working. Small caveat: does not work on Emacs ansi-term currently due to the use of `ESC[nG`, maybe others. See commit for more details.

mafredri:
@Hotschke, @sevanteri: are you guys still using pure with sshfs? I'd be happy if you could give this branch a spin, without disabling vcs_info on sshfs mounts :)!

mafredri:
Whoa, turns out we should be able to greatly simplify the preprompt update method... https://github.com/mafredri/pure/commit/24e3a3a7c9ae04be05fdc761bdec9e281e549135. I remember trying out newlines in the prompt before, but I couldn't get it working. Dunno what to say, seems to work better than the ansi escape fiddling.

mafredri:
Rebased onto latest master.

Hilzu:
@sindresorhus asked me to try this so I did. Have to say that it feels worse than before. I can't detect any latency when switching repositories without this so I can't really see any speed benefits.

Now when changing to a different repository with unpulled changes the down arrow appears for a while and then disappears after a second. This seems like a bug.

mafredri:
> Have to say that it feels worse than before. I can't detect any latency when switching repositories without this so I can't really see any speed benefits.

Thanks for the honest feedback. If you can explain, I would love to hear what exactly makes it feel worse.

This is indeed what worries me most with this PR. Pure is already so fast that on most machines the time it takes to check the git info is not easily noticeable to the human eye.

> Now when changing to a different repository with unpulled changes the down arrow appears for a while and then disappears after a second. This seems like a bug.

Weird, have not seen this. Would you mind providing e.g. an [asciinema](http://asciinema.org) recording?

---

Since `vcs_info` does a bit of unnecessary work, an alternative approach to this PR would be to stop using it in favor of executing the commands manually (e.g. to check if in git repo and the branch/status).

Hilzu:
My feeling about it being worse was completely based on the fiddly pull arrow. Now I'm unable to reproduce the issue anymore with this branch ¯\_(ツ)_/¯

mafredri:
Thanks @Hilzu for clarifying and trying to reproduce. I think I found the reason you encountered the issue. An edge case where a git fetch fails (for whatever reason) and is interpreted as a failed check for changes to upstream. Should be fixed in 5ebb0e2.

mafredri:
I've been noticing some visual glitches with this branch whenever I type before the prompt is ready, I don't notice the same with 9496e07 so I decided to merge it in (and rebase on master).

sindresorhus:
I've been running this for a couple of days now and seems to work fine.

I'd say let's merge this now. Mainly for https://github.com/sindresorhus/pure/pull/273/commits/9496e072f693ca40ffbe89de7a895885dae9a969.

mafredri:
Thanks for testing, and I agree. The simplified prompt update logic should even fix some old issues that are still open.

The git-parent-folder detection could surely still be improved, but the benefit would be minimal. Moving for merge now.

forivall:
yay, i got my fork working with this change https://github.com/sindresorhus/pure/compare/master...forivall:underline-repo-name

mafredri:
Nice @forivall, glad you were able to patch it in 😄!

Lugoues:
Has any thought been given on how to bring back the option allowing users to use vcs_info to customize the prompt?

mafredri:
@Lugoues not really, changing Pure through `vcs_info` configuration was never really a feature, just a side-effect of using it. What changes would you make?

Lugoues:
@mafredri Just various color tweaks and bold/italics based on branch or local repo state. Nothing too special. I didn't realize that using vcs_info wasn't really a feature Pure supported though, so no worries.

forivall:
@Lugoues Yeah, the recommended solution for customizations is to maintain your own fork, which, unfortunately for those of us who wants customization, shifts the burden of maintainership to us rather than the project maintainers. But they've often been willing to help with solving problems on customizations (as long as you don't ask them to maintain it 😛).

This is what I do for my fork where I make the name of the repo bold: https://github.com/forivall/pure/tree/underline-repo-name

(speaking of which, I need to submit a PR upstream for just this commit https://github.com/forivall/pure/commit/2848afb91cc71956914b8c11dc9b32a396a59379 )

mafredri:
@Lugoues there is one way to do that still, without maintaining your own fork, but it gets a bit hairy...

```zsh
# Rename the prompt_pure_preprompt_render function to prompt_pure_preprompt_render_orig.
typeset prompt_pure_preprompt_render_def=$(whence -f prompt_pure_preprompt_render)
eval ${prompt_pure_preprompt_render_def/prompt_pure_preprompt_render/prompt_pure_preprompt_render_orig}

# Hijack the pure renderer function.
prompt_pure_preprompt_render() {
	local cur_branch=$prompt_pure_vcs_info[branch]

	# If branch is set and we haven't already modified it...
	if [[ -n $prompt_pure_vcs_info[branch] ]] && [[ $cur_branch[1] != '%' ]]; then
		# Set branch color to yellow.
		prompt_pure_vcs_info[branch]="%F{2}${cur_branch}%f"
	fi

	# Call the original renderer.
	prompt_pure_preprompt_render_orig $@
}
```

You could drop this code anywhere after `prompt pure` and that would change the current git branch to yellow, as an example.

Takes advantage of the fact that the renderer is called anytime the prompt is updated, so we can do some preprocessing. Now, this probably goes without saying, but this could break at any time, so be warned 😅.

Lugoues:
@mafredri What are your thoughts on changing `prompt_pure_async_vcs_info` to have a call out to a custom function that a user, if they were so inclined, could create to modify their zstyle?

Something like this?
```
customize_vcs_info() {
  zstyle ':vcs_info:git*' formats " `tput sitm`%b`tput ritm`" "x%R"
}
```
```
prompt_pure_async_vcs_info() {
	setopt localoptions noshwordsplit
	builtin cd -q $1 2>/dev/null


	# configure vcs_info inside async task, this frees up vcs_info
	# to be used or configured as the user pleases.
	# only export two msg variables from vcs_info
	zstyle ':vcs_info:*' max-exports 2
	# export branch (%b) and git toplevel (%R)
	zstyle ':vcs_info:git*' formats '%b' '%R'
	zstyle ':vcs_info:git*' actionformats '%b|%a' '%R'

	whence -w customize_vcs_info > /dev/null && customize_vcs_info

	vcs_info

	local -A info
	info[top]=$vcs_info_msg_1_
	info[branch]=$vcs_info_msg_0_

	print -r - ${(@kvq)info}
}
```

mafredri:
@Lugoues while I do want to make Pure more easily hackable, I dislike that style of customization. I would see it as an intentional feature and it would have to be documented and maintained.

Furthermore, since this is an async task there is a timing aspect to it as well. The called function would have to be defined by the time the pure async worker is started, otherwise it will be unreachable. The user also can't rely on global shell state (e.g. setting a variable later) to change behavior due to this fact. It would be hard to properly document this behavior.

The easiest approach would simply be to copy-paste the entire `prompt_pure_async_vcs_info` function right after `prompt pure` and do the modifications you want (it would replace the one defined in Pure).

We could also split part of it into a separate function which you could override, but this gives very little value IMO. What I mean is something like this:

```shell
prompt_pure_vcs_info() {
	# configure vcs_info inside async task, this frees up vcs_info
	# to be used or configured as the user pleases.
	# only export two msg variables from vcs_info
	zstyle ':vcs_info:*' max-exports 2
	# export branch (%b) and git toplevel (%R)
	zstyle ':vcs_info:git*' formats '%b' '%R'
	zstyle ':vcs_info:git*' actionformats '%b|%a' '%R'

	vcs_info
}

prompt_pure_async_vcs_info() {
	setopt localoptions noshwordsplit
	builtin cd -q $1 2>/dev/null

	prompt_pure_vcs_info

	local -A info
	info[top]=$vcs_info_msg_1_
	info[branch]=$vcs_info_msg_0_

	print -r - ${(@kvq)info}
}
```

But the problem persist that if you make any kind of changes to Pure's vcs_info configuration, it could break in any future update of Pure and result in weird bugs. In this sense I believe it's better to be fully aware of what you're doing (e.g. by being forced to replace entire functions, etc.).

Lugoues:
@mafredri I see, I guess that makes some sense. I did notice in #306 there is talk of adding color support using zstyle, could this approach be taken to setup custom vcs_info settings too? Presumably the same issues apply there as well, no?

PR #274, opened 2017-01-18
Evaluate $PROMPT at render time

mafredri:
This allows `$PURE_PROMPT_SYMBOL` to be modified at any point.

screenshot 2017-01-18 18 09 26

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -368,7 +368,7 @@ prompt_pure_setup() {
 	[[ $UID -eq 0 ]] && prompt_pure_username=' %F{white}%n%f%F{242}@%m%f'
 
 	# prompt turns red if the previous command didn't exit with 0
-	PROMPT="%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f "
+	PROMPT='%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
 }
 
 prompt_pure_setup "$@"


sindresorhus:
Will this have any performance impact?

mafredri:
Nope, simple variable expansion is negligible.

sindresorhus:
What problem does this solve?

mafredri:
Slightly nicer for the user and more uniform behaviour across configuration options (the other options are evaluated when they are used as well)

Previously:

```zsh
PURE_PROMPT_SYMBOL='>'
prompt pure
```

Can now be done as:

```zsh
prompt pure
PURE_PROMPT_SYMBOL='>'
```

hkupty:
This PR broke my prompt. It now doesn't switch between `PURE_PROMPT_SYMBOL` or default, leaving a my prompt with the unrendered string:
`${PURE_PROMPT_SYMBOL:-❯}`

mafredri:
Sorry about that, although I do think this breakage could be due to a configuration mismatch. Do you, anywhere in your zshrc config, do `setopt NO_PROMPT_SUBST` or `unsetopt PROMPT_SUBST`? Either in lower case or without underscores?

Does `setopt PROMPT_SUBST` in your terminal fix the issue?

This is a configuration option pure already tries to set by default here https://github.com/sindresorhus/pure/blob/master/pure.zsh#L335

hkupty:
> Do you, anywhere in your zshrc config, do setopt NO_PROMPT_SUBST or unsetopt PROMPT_SUBST?
My `.zshrc` is very thin and I didn't set this config.

> Does setopt PROMPT_SUBST in your terminal fix the issue?
Yep, just fixed here.

Can this by any means be set by pure if not explicitly set/unset?

mafredri:
Good to know, thanks for reporting and testing. Completely forgot about these kinds of problems. Will look at it tomorrow.

hkupty:
No problem!

Thank you very much for helping..
I'll keep `PROMPT_SUBST` set on my `.zshrc` until this is figured.

Cheers

PR #275, opened 2017-01-18
Add indication when $SHLVL is > 1

mafredri:
Sometimes I find myself wondering if I'm still inside a new instance of a shell or not, as indicated by `$SHLVL`.

If this is something we want to merge, the readme would need to be updated as well (I can do that).

Here's an example of how this would look:
screenshot 2017-01-18 18 03 38

Thoughts @sindresorhus?

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -136,8 +136,13 @@ prompt_pure_preprompt_render() {
 	local git_color=242
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
-	# construct preprompt, beginning with path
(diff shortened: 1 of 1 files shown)


sindresorhus:
I'm neutral. I don't think I've ever been in a nested shell, but if you find it useful, feel free to add it.

mafredri:
Thanks, I'll give it some thought. I might be biased because I've been working on `zsh-async` quite a bit recently where this would've come in handy 😝 .

mafredri:
I'm not really finding this useful at the moment, and I wasn't sold on the look. Can't be bothered to make it look better so I'm closing this.

PR #281, opened 2017-01-25
Add option to show system time in prompt

johanneserber:
Hello,

I would like to display a system time in front of the prompt symbol like so:

screen shot 2017-01-25 at 14 24 28

The only way I found to do this was to edit the `PROMPT` value in the _pure.zsh_. What do you think about adding an option to do this?

If you want to keep your project free from such configurations I fully understand. Would there be another way to customise the prompt and keep up to date with your repo?

Best,
Johannes

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -390,8 +390,14 @@ prompt_pure_setup() {
 	# show username@host if root, with username in white
 	[[ $UID -eq 0 ]] && prompt_pure_username=' %F{white}%n%f%F{242}@%m%f'
 
+	# initialize prompt
+	PROMPT=''
+
+	# prepend system time
+	[[ "$PURE_PROMPT_SYSTEM_TIME" != '' ]] && PROMPT+='%F{white}%* '
+
 	# prompt turns red if the previous command didn't exit with 0
-	PROMPT='%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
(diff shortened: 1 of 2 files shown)


mafredri:
Thanks for the PR, however I think this is a bit too specific (in terms of preference) with little added value in Pure (e.g. how many people this would benefit from this).

The biggest problem here is the addition of a new configuration option (which you already considered), we would like to keep Pure as slim (in terms of options) as possible. Luckily, there are **currently** two ways to accomplish this without modifying the Pure code.

1. Set `PROMPT='%F{white}%* '$PROMPT` after pure has been initialised (via `prompt pure`)
2. Use the right prompt (`RPROMPT='%F{white}%*'`)

I emphasized *currently*  since option 1 might not be viable assuming a future feature would require modifying the `PROMPT`-variable at run-time.

sindresorhus:
I agree with @mafredri, but I would encourage you to add a recipe to the wiki when you find a good solution from the above suggestions: https://github.com/sindresorhus/pure/wiki :)

Closing this, but feel free to continue the discussion.

johanneserber:
Thanks for the quick response.

`1.` does the trick quite nicely without modifying the actual code. Perfect, thanks!

johanneserber:
@sindresorhus Added https://github.com/sindresorhus/pure/wiki#show-system-time-in-prompt. Thanks for this great repo. 🙂

PR #292, opened 2017-02-12
Add ability to change prompt colors

Kutsan:
It would be great.


mafredri:
Thanks for the PR, unfortunately this is a duplicate of #256, see discussion.

Kutsan:
I also respect your thoughts but I don't think a few color option makes clutter. I already changed a bunch of colors and merge them every new update.

mikeerickson:
@Kutsan @mafredri  I too would like alternate colors (I would like to see green vs magenta for success) but I can appreciate why there is a difference.

@Kutsan would you mind sharing your "alternate" settings?   I am trying to adjust color of the git branch as it is difficult for me to see with my vision issues.

Issue #167, opened 2015-10-18
Customizable dirty git symbol


justmytwospence:
Would be nice

sindresorhus:
Why? What would you use instead?

justmytwospence:
Probably some nice unicode symbol.

forivall:
this emoji: :imp: ?

justmytwospence:
An exclamation mark, an X, any number of things as per people's tastes. If the prompt char is deemed worthy of customizing, I thought this char might also. No problem closing this if you disagree.

mafredri:
The reason the unicode symbols are configurable is due to compatibility issues (see #139 for example). I can't say I have a solid stance on the issue, on the one hand I wouldn't want to introduce new options to pure, on the other hand, the dirty symbol is now the odd man out as the only one that is not configurable.

mafredri:
Instead of leaving this issue open in perpetuity, I'm closing this with the following argument:

The `*`-character is not a valid branch name in git and can thus never have any other meaning in `pure` than indicating a dirty repo. Due to it's placement, the asterisk also increases visibility and is generally a pleasant character to look at. Finally, it's supported on all systems and fonts (I hope) so there really is no reason to provide a way to change it (as we had to with our other unicode characters).

With that, I'm sorry @justmytwospence, I hope you can grow to love the asterisk :)!

forivall:
🙃

bentomas:
I'm okay with the decision you've made, and am not trying to be difficult with this comment, and I will just have to maintain my own fork of this project so I can change the 1 character in the codebase, but I have to admit, I'm pretty baffled by this decision.  Respectfully, it seems to me the lines you're drawing are pretty arbitrary.  

Why is someone's technical compatibility issues any different than @justmytwospence's and my aesthetic compatibility issues?  Your stated goals are "pretty, minimal and fast", and with this decision you've made it clear that you are going to define what is "pretty" (asterisk is pleasant to look at) and "minimal" (no new options), so why are you even bothering to support systems that don't have those characters?  Why aren't you requiring those people to switch their font choice?

It seems to me you would either say, pure will look and work like this, without options, OR, pure will have limited options and those options will be intuitive and consistent (like you can choose any of the symbols). But instead you've said, lets have limited options in inconsistent ways, so some people can make aesthetic choices about how pure looks, but others can't.

I honestly don't know what the right answer is.  I can certainly see how adding more options can be a slippery slope.

[...]

PR #293, opened 2017-02-12
Change * to ᵈ for clarity

originalpositron:
Basically what it says on the can: switching out the git dirty symbol ( ᵈ instead of *) should improve clarity about why symbol 'x' has appeared next to the git branch name in the preprompt. It also looks fancier without introducing extra visual distraction :)

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -316,7 +316,7 @@ prompt_pure_async_callback() {
 			if (( code == 0 )); then
 				prompt_pure_git_dirty=
 			else
-				prompt_pure_git_dirty="*"
+				prompt_pure_git_dirty="ᵈ"
 			fi
 
 			[[ $prev_dirty != $prompt_pure_git_dirty ]] && prompt_pure_preprompt_render


mafredri:
I've never seen anyone ask why `*` is beside the branch name, it's demonstrated in the screenshot and explained very early in the readme.

My feeling is that this doesn't really improve clarity, it's just different. And in my eyes the lone `d` looks kind of out-of-place.

I've outlined here https://github.com/sindresorhus/pure/issues/167#issuecomment-230821881 why I think `*` is the optimal character for the job. While reading this PR on my phone I also noticed the font did not support the character, e.g. we would have to provide configuration options for changing this due to not all fonts supporting it.

As an example, this would never be possible with `*`:

screenshot 2017-02-12 21 32 37

All-in-all, I think it's a fine idea, but not something we want to change in pure.

originalpositron:
Cool, thanks for the detailed reply!

mafredri:
Thanks for understanding!

PR #296, opened 2017-02-18
Helpful hint

shahzeb1:
A helpful hint which makes getting the active (non-comment) components from the .zshrc file much easier.


sindresorhus:
Not sure about this. This seems to indicate that the user can just use the command and post their .zshrc. The point we're trying to make is that you should reduce it as much as possible, not just removing comments.

shahzeb1:
Totally understood.

PR #302, opened 2017-03-10
Support configuration of git status timeout via PURE_GIT_STATUS_TIMEOUT and default to 5 seconds.

abutcher-gh:
A large(ish) repository I work with regularly usually returns status within 2 seconds.  But I've found that sometimes under load it takes a bit longer.  This patch allows for configuring the timeout before a repo is considered "large" to avoid an outlier status moving it into PURE_GIT_DELAY_DIRTY_CHECK territory.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -368,7 +368,7 @@ prompt_pure_async_callback() {
 			# When prompt_pure_git_last_dirty_check_timestamp is set, the git info is displayed in a different color.
 			# To distinguish between a "fresh" and a "cached" result, the preprompt is rendered before setting this
 			# variable. Thus, only upon next rendering of the preprompt will the result appear in a different color.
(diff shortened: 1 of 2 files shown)


mafredri:
Sorry for the delay @abutcher-gh, and thanks for your PR!

As you've observed, 2 seconds isn't really a magic bullet, I know some other users have had problems with this as well.

This is actually a feature I would like to remove, in favour of disabling dirty checking on a per-directory basis, for this reason I would like to avoid introducing the new configuration option. Do you think it would be OK to simply increase the limit to 5 seconds instead of making it configurable for now?

sindresorhus:
Yeah, let's go with 5 seconds. We already have too many options.

sindresorhus:
@mafredri What should we do about this one?

mafredri:
I suggest we close it, I can push the 2 -> 5 second change and reference this PR.

PR #304, opened 2017-03-17
Only expand the preprompt once

belak:
This came up because of this: https://github.com/njhartwell/pw3nage

I'd be open to cleaner ways of doing this, but this should fix the issue.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -131,19 +131,22 @@ prompt_pure_preprompt_render() {
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# construct preprompt, beginning with path
-	local preprompt="%F{blue}%~%f"
+	local preprompt='%F{blue}%~%f'
 	# git info
-	preprompt+="%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f"
+	preprompt+='%F{$git_color}${vcs_info_msg_0_}${prompt_pure_git_dirty}%f'
 	# git pull/push arrows
-	preprompt+="%F{cyan}${prompt_pure_git_arrows}%f"
+	preprompt+='%F{cyan}${prompt_pure_git_arrows}%f'
 	# username and machine if applicable
-	preprompt+=$prompt_pure_username
+	preprompt+='$prompt_pure_username'
 	# execution time
-	preprompt+="%F{yellow}${prompt_pure_cmd_exec_time}%f"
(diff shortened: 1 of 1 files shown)


mafredri:
Thanks for the PR, but you don't think the [current protection](https://github.com/sindresorhus/pure/blob/291574e652c18f3513524be88913a96038e93e19/pure.zsh#L112-L113) is sufficient?

Am I missing something, because to me it seems like this PR is doing exactly the same thing as what was done previously, except with an additional `$(print -Pn ...)`?

I do take security seriously though, so if you think there's something specific that must be fixed, I'm all ears, at least it doesn't seem like we're vulnerable to `pw3nage` 😆.

One other approach we could take is to (like you did in your PR) use single quotes and allow `prompt_subst` to be enabled for the preprompt. This avoids having to `$(print ...)`. But I don't see any benefit.

belak:
Unfortunately the current protection isn't sufficient because the double quotes currently get the values of the variables (including one containing the branch name) into the preprompt variable, then `print -P` is used to expand it.

I used the additional print for the else clause there. It uses the preprompt variable and checks the value against what was previously displayed. This needs to be an expanded version or that won't work as expected.

mafredri:
If you managed to break the Pure prompt I'd be interested in how. The `setopt no_prompt_subst` prevents variables from being expanded in the prompt, e.g. even if `$vcs_info_msg_0_` contains `$(fork bomb)` the prompt will just print out the text: `$(fork bomb)`.

belak:
Hm, I'm trying to reproduce this without prezto and haven't been able to yet... I'll get back to you if I manage to .

belak:
Alright, it was definitely my mistake. The agnoster theme adds something using add-zsh-hook and it looks like that was still firing after I switched to the pure prompt.

mafredri:
Aha, I managed to get hacked during by the Chinese!

However, the culprit was not as it seems: [pure.zsh#L183-L191](https://github.com/sindresorhus/pure/blob/291574e652c18f3513524be88913a96038e93e19/pure.zsh#L183-L191).

We we're actually only vulnerable during a preprompt update, specifically due to the prompt expansion (via `(S%%)` on the last line).

I guess the fix is a one-liner in this case, add `setopt no_prompt_subst` before the last expansion.

If you want to submit a new PR with it, I'd be happy to accept 😄, and thanks for bringing this to our attention.

belak:
Oh, or there is an actual issue. >_<

belak:
Ok, I managed to reproduce it... if the repo (pw3naga) is dirty, then it appears to trigger.

PR #305, opened 2017-03-17
Ensure no_prompt_subst is set when the expanded prompt is stored

belak:
This is an alternate fix as requested for the exploit mentioned in #304


belak:
CC @mafredri

mafredri:
Perfect, thanks =)!

PR #308, opened 2017-03-19
Add status indicator for Vi mode

gmathews:
This is for when a user has `bindkey -v` set to use the vi line editing mode. If they have it set, it would be nice to let the user know when they are in command mode as opposed to insert mode.
E.G.:
screen shot 2017-03-18 at 7 33 07 pm

Pressing escape toggles between normal(command)/insert mode in real time and the red N updates with it.

This will only show up if you are in command mode, which will only happen when you have enabled vi editing.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -105,6 +105,16 @@ prompt_pure_string_length_to_var() {
 	typeset -g "${var}"="${length}"
 }
 
+function zle-line-init zle-keymap-select {
+	# store our vi mode
+	if [[ "${KEYMAP}" == 'vicmd' ]]; then
+		prompt_pure_vi_mode="%F{red} N %f"
+	else
+		prompt_pure_vi_mode=""
+	fi
+	prompt_pure_preprompt_render
+}
+
 prompt_pure_preprompt_render() {
 	# store the current prompt_subst setting so that it can be restored later
 	local prompt_subst_status=$options[prompt_subst]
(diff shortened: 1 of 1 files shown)


mafredri:
Thanks for the PR. I'm not really opposed to supporting VIM-mode out of the box, but I'm not ready to accept this PR as-is.

One big issue here is that we're potentially overwriting the users `zle-line-init` and `zle-keymap-select` functions. To fix avoid this, there are two possible solutions:

1. Check if they are already bound, if bound skip, if not, bind (like we do with clear screen)
2. If bound, wrap the users widget (e.g. like `zsh-syntax-highlighting` does)

Option 1 not optimal, since it's a bit hit or miss whether or not this feature is enabled, and option 2 adds extra complexity.

PS. Have you seen the [example](https://github.com/sindresorhus/pure/wiki#show-current-mode-when-using-vi-keybindings) in the Wiki? I haven't really used VIM-mode much, but I like that it's discrete.

sindresorhus:
@mafredri What should we do about this?

mafredri:
@sindresorhus I think we can close this for now? I see there is some support for this PR, though. So supporting VIM-mode out of the box might be worth considering.

Since [add-zle-hook-widget](https://github.com/zsh-users/zsh/blob/master/Functions/Misc/add-zle-hook-widget) was added in zsh 5.3, we could use it for hooking into `zle-line-init` and `zle-keymap-select`. Older versions like 5.2 would get no official support for VIM-mode.

Question I think is, do we implement something similar to this PR or the one in the Wiki? Personally I like the Wiki-version since it doesn't introduce new UI components. However, I think we have to add an option for the symbol if we implement it 😞.

sindresorhus:
> Since [add-zle-hook-widget](https://github.com/zsh-users/zsh/blob/master/Functions/Misc/add-zle-hook-widget) was added in zsh 5.3, we could use it for hooking into `zle-line-init` and `zle-keymap-select`. Older versions like 5.2 would get no official support for VIM-mode.

👍 

> Question I think is, do we implement something similar to this PR or the one in the Wiki? Personally I like the Wiki-version since it doesn't introduce new UI components. However, I think we have to add an option for the symbol if we implement it 😞.

I'm not really sure how they differ, but I trust your judgement, so up to you.

Issue #320, opened 2017-05-03
Virtualenv name no longer shown in prompt


rif:
### General information

- Pure version: 4cdd0cf
- ZSH version: 5.3.1
- Terminal program & version: rxvt-unicode 9.22, sakura 3.3.4
- Operating system: arch linux
- ZSH framework: zgen

I have:
- [ x] Tested with another terminal program and can reproduce the issue: 
- [] Followed the [Integration](https://github.com/sindresorhus/pure#integration) instructions for my framework

### Problem description
The python virtualenv name used to be printed in the prompt after the activation using virtualenvwarapper, not it stopped working

### Reproduction steps

1. create a virtualenv
2. activate it
3. virtualenv name should appear in prompt

### My `.zshrc`:
```shell
zgen load mafredri/zsh-async
zgen load sindresorhus/pure
```

mafredri:
Pure has never officially supported virtualenvs (e.g. #254) but it has worked since Pure used to handle the preprompt differently, now the preprompt is part of `PROMPT` (or `PS1`) which means the virtualenv is being set, but is cut off afterwards:

For example, this is what the prompt looks like after sourcing the virtualenv:

```
PS1=$'(my_virtualenv) \n%{\C-M%}%F{blue}%~%f %F{242}${prompt_pure_vcs_info[branch]}${prompt_pure_git_dirty}%f %F{cyan}${prompt_pure_git_arrows}%f\n%{\C-M%}%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
```

But what happens afterwards is that Pure cuts away the excess (everything before the last newline):

```
PS1=$'(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
```

And then adds a new, clean preprompt.

I'll try to do something about this soon, in the meanwhile, here's a patch you could use that should bring back old behavior:

```patch
diff --git a/pure.zsh b/pure.zsh
index 78e048f..11adf7c 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -178,6 +178,9 @@ prompt_pure_precmd() {
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks
 
+	psvar[12]=
+	[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+
 	# print the preprompt
 	prompt_pure_preprompt_render "precmd"
 
[...]

PR #321, opened 2017-05-03
Restore old virtualenv behavior by only modifying preprompt

mafredri:
It seems 4cdd0cf4f48d0c2064700c4fc60083bc0b3f6693 prevents some previous use-cases (e.g. with regard to virtualenvs: #320).

~~This PR restricts prompt modifications to only what is present in the preprompt, by surrounding the preprompt in non-existent variables (that prints empty). I'm not perfectly happy with this approach as it feels a bit hackish, but it should be good enough for now.~~ Simplified in d8322b7.

There is an alternative way to go about this as well, as proposed in the issue https://github.com/sindresorhus/pure/issues/320#issuecomment-298942782. However, that approach is not as versatile as this PR. This PR should fix all possible cases of appending/prepending to prompt by other shell scripts.

Even if this PR is merged, we might still want to implement the patch from #320 once we properly decide on how to go forward with virtualenvs.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -138,21 +138,23 @@ prompt_pure_preprompt_render() {
 	# Execution time.
 	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{yellow}${prompt_pure_cmd_exec_time}%f')
 
-	# Cleanup prompt (remove preprompt).
-	local trim_preprompt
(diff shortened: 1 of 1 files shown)


mafredri:
I'm going to merge this in now as I'm fairly confident it shouldn't introduce new issues, want to get this fix out ASAP.

PR #322, opened 2017-05-05
Remove deprecated clear-screen ZLE widget

mafredri:
This is no longer needed because `$PROMPT` contains the entire prompt (including preprompt) since 4cdd0cf4f48d0c2064700c4fc60083bc0b3f6693.

Mention in readme FAQ was also removed.


sindresorhus:
Great! Always nice to be able to remove a caveat :)

PR #323, opened 2017-05-07
Improve the style of the continuation prompt (PS2)

mafredri:
I've always thought the PS2 looked a bit out-of-place in Pure and I wanted to improve it, so I gave it a whack.

This PR does two things:

* Stylize the PS2 prompt with colors and the Pure prompt symbol `❯` (also includes the exit status in the PS2 prompt)
* Prevent prompt updates when PS2 is displayed (it's not possible to update PS1 with `reset-prompt` at this point)

This is what it looks like now:

screenshot 2017-05-07 15 08 24

A variation I tried was to include `❯` in front as well:

screenshot 2017-05-07 15 06 25

But while looking OK was just not quite there.

I also looked at a bunch of different arrows and tried them out but couldn't find a good fit, closest ones were `↳` (but it did not quite align in the middle of the text, thus looking misaligned) and `⥅`. Neither of which I felt added value.

### Considerations

* Do we need to consider how this looks when virtualenvs (#254) are activated if we were to show them in the same manner?
* We could limit the number of entries (`cmdand`, etc.) by using `%N_` where `N` is the max number of entries, but I can't think of a very good reason to do this, at least it has never been an issue for me

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -642,6 +642,7 @@ prompt_pure_setup() {
 
 	prompt_pure_state_setup
 
+	zle -N prompt-pure-reset-prompt prompt_pure_reset_prompt
 	zle -N prompt_pure_update_vim_prompt_widget
 	zle -N prompt_pure_reset_vim_prompt_widget
 	if (( $+functions[add-zle-hook-widget] )); then
--- a/pure.zsh
+++ b/pure.zsh
@@ -651,11 +651,12 @@ prompt_pure_setup() {
 
 	# if a virtualenv is activated, display it in grey
 	PROMPT='%(12V.%F{242}%12v%f .)'
-	PROMPT2='%(12V.%F{242}%12v%f .)'
 
 	# prompt turns red if the previous command didn't exit with 0
 	PROMPT+='%(?.%F{magenta}.%F{red})${prompt_pure_state[prompt]}%f '
-	PROMPT2+='%F{242}%_%f %(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
+
+	# Indicate continuation prompt by ... and use a darker color for it.
+	PROMPT2='%F{242}... %(1_.%_ .%_)%f%(?.%F{magenta}.%F{red})${prompt_pure_state[prompt]}%f '
 
 	# Store prompt expansion symbols for in-place expansion via (%). For
 	# some reason it does not work without storing them in a variable first.
--- a/pure.zsh
+++ b/pure.zsh
@@ -103,16 +103,6 @@ prompt_pure_preexec() {
 	export VIRTUAL_ENV_DISABLE_PROMPT=${VIRTUAL_ENV_DISABLE_PROMPT:-12}
 }
 
-prompt_pure_reset_prompt() {
-	if [[ $CONTEXT = cont ]]; then
-		# When the context is cont, PS2 is active and calling
-		# reset-prompt will have no effect on PS1.
-		return
-	fi
-
-	zle .reset-prompt
-}
-
 prompt_pure_preprompt_render() {
 	setopt localoptions noshwordsplit
 
@@ -170,7 +160,7 @@ prompt_pure_preprompt_render() {
 		print
 	elif [[ $prompt_pure_last_prompt != $expanded_prompt ]]; then
 		# Redraw the prompt.
-		zle && zle prompt-pure-reset-prompt
+		prompt_pure_reset_prompt
 	fi
 
 	typeset -g prompt_pure_last_prompt=$expanded_prompt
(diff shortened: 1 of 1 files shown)


sindresorhus:
Do we have to display `cmdand`? And I would go for a different color for the prompt character when in the continuation prompt, to signify it's a continuation. Maybe gray or dim cyan.

mafredri:
> Do we have to display `cmdand`?

I don't see a way around it, I think it's useful/important information. E.g. tells you why you ended up on a continuation prompt (might not always be obvious). Personally I wanted to try replacing it it with e.g. `&&`, but as far as I understand the (zsh) source, we cannot read the value of `%_` that would be printed to the terminal since it's done on-the-fly during prompt rendering/expansion and the only time the "execution context" is correct, is then.

> And I would go for a different color for the prompt character when in the continuation prompt, to signify it's a continuation. Maybe gray or dim cyan.

My thought was that `cmdand` was enough of an indication. Keeping the behavior of PS1 with regard to color was a simple solution, that IMO looks good. Not sure what a good color would be, if we go with `242` (same as `cmdand`) then my feeling is it doesn't really look like a prompt indication anymore, `cyan` on the other hand doesn't really change much (for me) compared to magenta:

screenshot 2017-05-07 18 49 53
screenshot 2017-05-07 18 49 48

mafredri:
Looks like there's one aspect I did not consider, when there isn't any content for `%_`, currently looks like this (with cyan mod):

screenshot 2017-05-07 18 55 25

mafredri:
So given the above, I think we definitely need to change the prompt color for continuations, or alternatively, use a different symbol (e.g. `❮` or `❯❯`). I'll try to experiment with some colors.

(Ignore the magenta color):

screenshot 2017-05-07 19 10 27

sindresorhus:
> Personally I wanted to try replacing it it with e.g. &&

That's what I was thinking too. `cmdand` doesn't look very good.

> So given the above, I think we definitely need to change the prompt color for continuations, or alternatively, use a different symbol (e.g. ❮ or ❯❯).

Do we need a symbol at all? The colors difference kinda makes the distinction.

Alternatively:

```
❯ echo -n Hello &&
cmdand └ foo
```


```
❯ echo -n Hello &&
cmdand │ foo
```

mortenvn:
I'm really like the new style! It fits right in with the rest of Pure. 👍

#### Some thoughts on the things you've already discussed

> Personally I wanted to try replacing it it with e.g. &&

Shame that this isn't possible. It would have been glorious! 

It took me a while to figure out that `cmdand` actually meant `&&` and that it wasn't some sort of bug or placeholder text. (I'm not a smart man 😩)

> Do we need a symbol at all?

In my opinion, the symbol is a great way to indicate that everything fine, that your still in the Pure prompt and that you can keep writing commands.

> I think we definitely need to change the prompt color for continuations, or alternatively, use a different symbol (e.g. ❮ or ❯❯).

Of the ideas presented, I really liked `❯❯`. It's simple, it looks like the Pure symbol, but at the same time looks different enough to be noticeable.

#### Delimiter choices (to support virtualenv)
As we discussed in https://github.com/sindresorhus/pure/pull/254, I tried to see how this could be combined with virtualenv support:

delimiter-type

As you can see I just smacked pipe symbols of different weights between them. I think nr. 2 (medium vertical bar) is the best one as the pipe has the same thickness as ❯.

### Color choices
colors
I prefer the first two, with a slight edge towards nr. 1. This is really subjective, but I just think it looks the cleanest and most consistent.

There is, however, a case to be made for the cyan one as well: When you have text (like "cmdand") I don't think the color matters that much since it remains easy to understand that you've entered _"Fuck it! We'll do it live!"_ mode. But in the first example (`echo Hello \`) we don't have that luxury. In this case the cyan color might help indicate the change. 

I'm not a big fan of making the ❯ symbol grey. Having three grey things (virtualenv, the text and the ❯) next to each other makes it harder to parse. I also feel that the color helps guide the eye and indicate that the prompt is ready for more.

### Alternative solution
I don't really know how successful this proposal is, but I was curious to see what could be done without the text:
alternative

In the end I think it's an interesting idea, but will probably cause some confusion. For example: Note how you go right from three prompts (❯❯❯) to being finished. Intuitively I would have expected to get back to two before it quit.

sunjon:
The last proposal(venv  + alt color) looks great. Is there something stopping it being able to drop down an indentation level on the first 'fi' ?

mafredri:
Thanks for the awesome feedback @mortenvn, you've given me a bit to think about, I'll try to get back to this in the coming days 😄.

> Shame that this isn't possible. It would have been glorious!

Yup 😢.

sindresorhus:
> Shame that this isn't possible. It would have been glorious!
> > Yup 😢.

Open a feature request on the ZSH issue tracker?

sindresorhus:
@mafredri How should we proceed with this?

mafredri:
@sindresorhus apologies for leaving this hanging, I'll try to get around to fixing this up during this week. I need to reacquaint myself with this again and consider @mortenvn suggestions 😄.

> Open a feature request on the ZSH issue tracker?

They don't really have an issue tracker, just the zsh-workers mailing list (and I dread mailing lists). Anywho, I'll try to get around to posting this on there, maybe we could have this in Zsh 5.4 if we're lucky.

mafredri:
Heading off to vacation now so I didn't get this quite done. I did a bit of experimenting and testing mortenvn's suggestions though, and I can't help but feel that the PS2 gets crowded, esp. when a virtualenv is active. An idea I had was to instead just show:

```
myvirtialenv ❯ echo hi &&
... andand ❯ echo bye
```

The `...` would indicate PS2 (continuation prompt).

Colors would be dark as above for everything before the `❯`.

sindresorhus:
@mafredri What should we do about this? Happy to go with the simpler solution.

mafredri:
I've rebased the branch and simplified the continuation prompt, it no longer displays virtualenv and uses `...` to signify that we have entered continuation mode.

Screen Shot 2019-06-11 at 23 04 20

image

I think this works good enough to be merged, thoughts?

PR #325, opened 2017-05-08
Add support for python virtualenv

mortenvn:
When a virtualenv is activated, the name of the virtualenv will be displayed right before the `❯` symbol.

screen shot 2017-05-08 at 20 42 24

**About the code**
I haven't really done anything fancy. I ended up just making a couple of tweaks to the patch @mafredri posted in [#320 (comment)](https://github.com/sindresorhus/pure/issues/320#issuecomment-298942782).

**P.S.** Any particular reason why you (@mafredri) chose the 12th position in `psvar`? I never really understood why, but I figured that there's probably a good reason 😛

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -176,6 +176,10 @@ prompt_pure_precmd() {
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks
 
+	# store name of virtualenv in psvar if activated
+	psvar[12]=
+	[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+
 	# print the preprompt
 	prompt_pure_preprompt_render "precmd"
 
@@ -425,6 +429,9 @@ prompt_pure_setup() {
 	# Prevent percentage showing up if output doesn't end with a newline.
 	export PROMPT_EOL_MARK=''
 
+	# disallow python virtualenvs from updating the prompt
+	export VIRTUAL_ENV_DISABLE_PROMPT=1
+
 	prompt_opts=(subst percent)
 
 	# borrowed from promptinit, sets the prompt options in case pure was not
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -453,8 +460,11 @@ prompt_pure_setup() {
 	# show username@host if root, with username in white
 	[[ $UID -eq 0 ]] && prompt_pure_username='%F{white}%n%f%F{242}@%m%f'
 
+	# if a virtualenv is activated, display it in grey
+	PROMPT='%(12V.%F{242}${psvar[12]}%f .)'
```
mafredri: I would prefer if we used `%12v` instead of `${psvar[12]}`. Mostly for consistency.

mafredri:
> P.S. Any particular reason why you (@mafredri) chose the 12th position in psvar? I never really understood why, but I figured that there's probably a good reason 😛

Haha. I guess `$VIRTUAL_ENV` is 12 characters, but mostly I just wanted to select a number high enough that it wouldn't collide with anything our users might be using 😅.

pure.zsh
mortenvn: I totally agree. Fixed :)

pure.zsh
```diff
@@ -176,6 +176,10 @@ prompt_pure_precmd() {
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks
 
+	# store name of virtualenv in psvar if activated
+	psvar[12]=
+	[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
```
mafredri: I had a thought: what if we check `[[ -n $VIRTUAL_ENV ]] && [[ $VIRTUAL_ENV_DISABLE_PROMPT = 1 ]]` here? My thought is that this would allow users to opt-out of us managing the virtualenv by e.g. `unset VIRTUAL_ENV_DISABLE_PROMPT`.

mafredri:
Although I wrote https://github.com/sindresorhus/pure/pull/325#discussion_r117045659, I'm not sure if it makes any sense at this point, so I went ahead and merged 😄 🎉.

Nifled:
Does this work out of the box, or is there something I have to modify? I haven't been able to see anything.

PR #328, opened 2017-06-16
Use glob instead of regex for dir matching

mafredri:
Discovered this issue by accident.

Sometimes directory names might contain characters that cause an invalid regex, for example, cd:ing into the directory `test[` would result in the following error:

```
prompt_pure_async_callback:21: failed to compile regex: brackets ([ ]) not balanced
```

We avoid this problem by using globbing instead of regex when checking if the path matches.

**Sidenote:** I'm a bit worried that there could be some zsh option (or combination of options) that causes this to misbehave, but I can't think of any.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -286,7 +286,7 @@ prompt_pure_async_tasks() {
 	typeset -gA prompt_pure_vcs_info
 
 	local -H MATCH
-	if ! [[ $PWD =~ ^$prompt_pure_vcs_info[pwd] ]]; then
+	if ! [[ $PWD = ${prompt_pure_vcs_info[pwd]}* ]]; then
 		# stop any running async jobs
 		async_flush_jobs "prompt_pure"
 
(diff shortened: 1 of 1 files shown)


jschaf:
For the worry about zsh options; you can force all options back to default values by adding `emulate -L zsh` to the beginning of an function.  The [user guide](http://zsh.sourceforge.net/Guide/zshguide02.html) has all the gory details.

mafredri:
Thanks for the suggestion @jschaf, I'm not actually that worried though 😛. I often rely on `emulate` for scripts, but in the case of Pure I prefer to make decisions that come from a place of understanding instead of fear. My hope is that people could learn something new about zsh or discover new features by reading the Pure source (not that I'm an expert!). That can't happen if we do things just to be on the safe side.

Now, if this PR were to break someones shell, it would be unfortunate, but I (and hopefully others) would learn something new.

PR #336, opened 2017-08-02
Config options and speed improvement

BrandonRoehl:
The speed improvement comes with rendering the prompt fully and not having logic execute latter given that the render function is called every time anyways this can reduce multiple logical calls.

Also added options for color configuration. The reason for this is to use 256bit colors instead for terminals that can support it.

```zsh
export PURE_PATH_COLOR=045
export PURE_GIT_ARROW_COLOR=87
export PURE_GIT_COLOR=245
export PURE_PROMPT_COLOR=177
export PURE_PROMPT_ERROR_COLOR=203
```


BrandonRoehl:
Removed unless prompt configuration options because PROMPT can be configured anyways
```zsh
export PURE_PATH_COLOR=045
export PURE_GIT_ARROW_COLOR=87
export PURE_GIT_COLOR=245
export PURE_TIME_COLOR=215
PROMPT='%(12V.%F{242}%12v%f .)%(?.%F{117}.%F{203})❯%f '
```

Issue #306, opened 2017-03-18
Proposal: configure Pure colors via prompt init


mafredri:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/pure/issues/306/badge.svg" />](https://issuehunt.io/r/sindresorhus/pure/issues/306)

We receive a lot of requests for changing colors in Pure, one prevalent reason is color blindness and hard-to-distinguish colors. Many of the [themes that ship with Zsh](https://github.com/zsh-users/zsh/tree/master/Functions/Prompts) allow color schemeability during prompt init and I believe we could utilize this in Pure as well.

#### Example:

```
Usage: prompt pure [<status_ok_color> [<status_error_color> [<path_color> [<git_color> [<arrow_color> [<duration_color> [<username_color> [<root_color>]]]]]]]]

prompt pure magenta red blue 242 cyan yellow 242 white
```

These could maybe even be just `color1`, `color2`, etc. 

#### Benefits:

- No backwards incompatibility
- No new environment options
- Allows changing all colors (or as many as we expose, maybe a more limited set would be better, to begin with)

#### Downsides:

- "Hard" to change just one color (previous colors must also be filled)
- Colors can't officially<sup>1</sup> be changed at runtime based on whatever reason, e.g. inside a read-only directory, so the user would want to change the directory prompt to `red`, or whatever...
- Do we need to document how each color affects the prompt?

<sup>1</sup> Unofficially the internals are probably such that the user could change them, but this would of course be unsupported and could break at any time.

---

As a bonus, we could implement `prompt_pure_help` to provide a help dialogue (on how to use/modify pure) for the user. This can be displayed with `prompt -h pure`.

#### References:

- Add support for custom prompt arrow color (#256)
- Feature request | arrow prompt ❯ color (#303)
- Others: #292, #155

// ping @sindresorhus

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/pure/issues/306).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/pure/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

sindresorhus:
Does ZSH have a nicer way to do options? I don't like positional arguments like this, and environment variables are global.

sindresorhus:
[...]

PR #337, opened 2017-08-03
Git States

BrandonRoehl:
screen shot 2017-08-03 at 11 56 05 am

# 4 major changes
1. Speed improvement with the `prompt_pure_preprompt_render` not passing expressions to the `PROMPT`
2. Config options for the prompt arrow removed because you can set `PROMPT` anyways
3. Config options for colors in the top added. Why? Because 256bit colors are nice for not changing the terminal setup. And forcing them kinda ousts people without 256bit support in zsh or tmux.
4. Git now shows the modified, untraced, and staged files denoted by a `*` `.` `+` respectively
5. Config `PURE_GIT_UNTRACKED_DIRTY` removed in favor of displaying the `.`

## My Config for 256bit colors
```zsh
export PURE_PATH_COLOR=045
export PURE_GIT_ARROW_COLOR=87
export PURE_GIT_COLOR=245
export PURE_TIME_COLOR=215
PROMPT='%(12V.%F{242}%12v%f .)%(?.%F{177}.%F{203})❯%f '
```

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -241,7 +241,7 @@ prompt_pure_async_git_dirty() {
 
 	local -A info
 
-    git ls-files --other --exclude-standard 2> /dev/null
+    test -z `git ls-files --other --exclude-standard`
     info[untracked]=$?
 
     git diff --quiet 2> /dev/null
--- a/pure.zsh
+++ b/pure.zsh
@@ -105,29 +105,29 @@ prompt_pure_preprompt_render() {
 
 	# Set color for git branch/dirty status, change color if dirty checking has
 	# been delayed.
-	local git_color=242
+	local git_color=${PURE_GIT_COLOR:-242}
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
 
 	# Initialize the preprompt array.
 	local -a preprompt_parts
 
 	# Set the path.
-	preprompt_parts+=('%F{blue}%~%f')
+	preprompt_parts+=("%F{${PURE_PATH_COLOR:-blue}}%~%b%f")
 
 	# Add git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
 	if [[ -n $prompt_pure_vcs_info[branch] ]]; then
-		preprompt_parts+=("%F{$git_color}"'${prompt_pure_vcs_info[branch]}${prompt_pure_git_dirty}%f')
+		preprompt_parts+=("%F{$git_color}${prompt_pure_vcs_info[branch]}${prompt_pure_git_dirty}%f")
 	fi
 	# Git pull/push arrows.
 	if [[ -n $prompt_pure_git_arrows ]]; then
-		preprompt_parts+=('%F{cyan}${prompt_pure_git_arrows}%f')
+		preprompt_parts+=("%F{${PURE_GIT_ARROW_COLOR:-cyan}}${prompt_pure_git_arrows}%f")
 	fi
 
 	# Username and machine, if applicable.
-	[[ -n $prompt_pure_username ]] && preprompt_parts+=('$prompt_pure_username')
+	[[ -n $prompt_pure_username ]] && preprompt_parts+=("$prompt_pure_username")
 	# Execution time.
-	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{yellow}${prompt_pure_cmd_exec_time}%f')
+	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=("%F{${PURE_TIME_COLOR:-yellow}}${prompt_pure_cmd_exec_time}%f")
 
 	local cleaned_ps1=$PROMPT
 	local -H MATCH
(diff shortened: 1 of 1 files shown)


BrandonRoehl:
The `*` `.` `+` symbols could be something completely different and in a different order too. This is just what I came up with

mafredri:
Hi, and thanks for the PR.

I would suggest splitting this up in separate PR's so each topic of change can be discussed in separation.

Here are my initial thoughts:

> 1. Speed improvement with the prompt_pure_preprompt_render not passing expressions to the PROMPT

How much of a speed improvement is this, did you perform any benchmarks?

Also, I believe this change might open us up to a security hole that can e.g. be triggered by [njhartwell/pw3nage](https://github.com/njhartwell/pw3nage), but I haven't tested the PR. Can you confirm if this is an issue or not?

> 2. Config options for the prompt arrow removed because you can set PROMPT anyways

This is a backwards-incompatible change, anyone relying on the environment variable would be affected. Although setting the prompt manually is possible, it's not very intuitive or obvious. It might even break in future, or new features might be missing.

> 3. Config options for colors in the top added. Why? Because 256bit colors are nice for not changing the terminal setup. And forcing them kinda ousts people without 256bit support in zsh or tmux.

Colors is something we're actively considering, however, we would like to avoid introducing new environment variables. This is currently being discussed in https://github.com/sindresorhus/pure/pull/256 and https://github.com/sindresorhus/pure/issues/306 (also https://github.com/sindresorhus/pure/pull/292). It would be great if we could see a PR for a `zstyle` approach for colors 😄.

> 4. Git now shows the modified, untraced, and staged files denoted by a * . + respectively

I'm not sure we want to go down this path with Pure, personally I don't see much value in showing that something is staged or untracked since an indicator is not granular enough. I use `git status` for that. But this would be better discussed in separation.

> 5. Config PURE_GIT_UNTRACKED_DIRTY removed in favor of displaying the .

Hmm, I think the reason this option exists (IIRC) is to speed up dirty checking for large repos (think chromium). Can you confirm @sindresorhus?

---

PS. We use tab-indentation instead of spaces in Pure, I would recommend installing the `editorconfig` plugin for your editor for future PR's 🙂.

sindresorhus:
> Hmm, I think the reason this option exists (IIRC) is to speed up dirty checking for large repos (think chromium). Can you confirm @sindresorhus?

Yup

sindresorhus:
And I agree with everything @mafredri said.

BrandonRoehl:
> Also, I believe this change might open us up to a security hole that can e.g. be triggered by njhartwell/pw3nage, but I haven't tested the PR. Can you confirm if this is an issue or not?

This is talking about examples where the `PROMPT` has the evaluating such as what you had
```zsh
PROMPT='${BRANCH}'
echo $PROMPT # > ${BRANCH}
# Evaluates BRANCH in every dir so if BRANCH has a resolvable path to a file can be executed
```
what I suggested
```zsh
PROMPT="${BRANCH}"
echo $PROMPT # > master
# even after dir change
# This is evaluated in our environment so unless the script we have has a function that is resolvable nothing happens
```
### Originally
Note every command to get evaluated takes time as well
screen shot 2017-08-12 at 11 41 44 am

### After
Only going to evaluate colors no `$`
screen shot 2017-08-12 at 11 39 18 am

Tested that it still works because you were setting the stuff anyways it was just changing from single quotes to double quotes. This is the really minor speed improvement because the prompt doesn't evaluate anything for values. The prompt is slower than in the pre command because you are already doing variable and register access.

The general idea is you should never set `PROMPT` in the pre command but if your doing it anyways why not.
The best speed is just set `PROMPT` once in setup and use the `psvar[]` to handle the few substitutions in it removing all `$` signs from your prompt and security issues.

BrandonRoehl:
I am working on a rewrite [BrandonRoehl/zsh-clean](https://github.com/BrandonRoehl/zsh-clean/blob/master/clean.zsh)

1. It sets `PROMPT` once
2. `psvar[]` is used
3. Because of this the `PROMPT` is 100% customizable after
4. It doesn't require external libraries such as `async` this is handled by zsh `&`
5. The `precmd` is 4 lines long
6. `vcs_info` is configured to use hooks allowing it to support `ALL` of the vcs that vcs_info allows while still supporting git unstaged and git arrows as an action format

The `precmd` is slightly slower because it lacks callbacks from async.

mafredri:
I didn't get a clear answer regarding pw3nage, so I tried it out myself and this PR is indeed vulnerable to it. Expanding the git branch name in-place, as in this PR, is not safe.

Original: `'$BRANCH'` -> render prompt -> prints: `$(./pw3n)`.

PR: `"$BRANCH" -> "$(./pw3n)" -> render prompt -> evaluates `$(./pw3n)` (e.g. runs the command).

> The prompt is slower than in the pre command because you are already doing variable and register access.

Do you have any facts to back this up? I would think that variable lookups are so cheap that it's not worth thinking about (at this scale).

mafredri:
> The general idea is you should never set PROMPT in the pre command but if your doing it anyways why not.

I've never heard of this, could you point me to some resource or provide background?

> The best speed is just set PROMPT once in setup and use the psvar[] to handle the few substitutions in it removing all $ signs from your prompt and security issues.

I'd also be interested to learn more of this, e.g. why would it be faster to use psvar? Do you know any resources?

BrandonRoehl:
> Do you have any facts to back this up? I would think that variable lookups are so cheap that it's not worth thinking about (at this scale).
Seeing that the pre render then allows that sting to be evaluated after I would agree on the fact that that the retrieval from memory would probably be worth it to keep it from rendering.

> I'd also be interested to learn more of this, e.g. why would it be faster to use psvar? Do you know any resources?

`psvar` isn't inherently any faster that doing a standard lookup of any other variables. What it offers is
`%v#` properly escaping and not evaluating anything. Setting the prompt once is what speeds it up a lot. Because then the big o is what decreases when you do less assignments and reads

https://www.zsh.org/mla/users/2005/msg00863.html

## Big O
1. check cmd time
2. unset timestamp
3. VCS_INFO (async)
4. set title
5. unset ps var 12
6. test for virtual environment
7. store virtual environment

pre render
8 test
6 assignments
+= is and assignment and lookup and you have 4 of them
and then joining ps1 into prompt every single time

This is the pre command trace

Where as if you assign prompt only once then all you have are variable lookups

it could be just

1. check cmd time
2. unset timestamp
3. VCS_INFO (async)
4. set title
5. assign psvar

## psvar

psvar can also test the existence of variables in prompt its own way too using `%(V1. %v1.)` to add a space before the v1

mafredri:
I'm not unfamiliar with `psvar`, I know what it does, but I can't find anything regarding prompt rendering speed in `zsh-users 9188` that you linked to. I would be very interested in some facts regarding prompt rendering speed when evaluating variables versus not (e.g. using `psvar`).

PS. I don't see how Big-O has anything to do with this. We're not really expressing the complexity of algorithms here, just a few assignments ;-).

PR #338, opened 2017-08-10
Add back the prompt_sp option for zsh >= 5.4.1

belak:
In 5.4.1, this option was reset between prompts, so to retain the
previous default behavior, this should be added back.

This was the change which caused it to be reset: https://github.com/zsh-users/zsh/commit/43e55a9bcd2c90124a751f2597d2f33cb6e3c042#diff-bb10d67e7a8561b66a53a805f3c77a40R233

This is the documentation on the option itself: http://zsh.sourceforge.net/Doc/Release/Options.html#Prompting

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -432,11 +432,11 @@ prompt_pure_setup() {
 	# disallow python virtualenvs from updating the prompt
 	export VIRTUAL_ENV_DISABLE_PROMPT=1
 
-	prompt_opts=(subst percent)
+	prompt_opts=(sp subst percent)
 
 	# borrowed from promptinit, sets the prompt options in case pure was not
 	# initialized via promptinit.
-	setopt noprompt{bang,cr,percent,subst} "prompt${^prompt_opts[@]}"
+	setopt noprompt{bang,cr,percent,sp,subst} "prompt${^prompt_opts[@]}"
 
 	if [[ -z $prompt_newline ]]; then
 		# This variable needs to be set, usually set by promptinit.


mafredri:
Thanks @belak, I see no problem with merging this PR, since `sp` has no effect as long as `cr` is not enabled. However, if a user has manually enabled `cr` (via `setopt`), then this change would suddenly start including whitespace at the end of lines that do not terminate with a newline, but this is minor.

Just wondering what the benefit of merging this is, sticking closer to zsh defaults?

mafredri:
> Just wondering what the benefit of merging this is, sticking closer to zsh defaults?

Or more importantly, any downsides on versions of zsh <5.4.1?

belak:
Maybe I've made a mistake... we were seeing issues in prezto prompts where things like `echo -n hello world` wouldn't display anything at all... and it doesn't look like this fixes the underlying problem in pure... since I don't see anything even with the "fix".

belak:
Ok, it looks like all the prompts I tested had `prompt_cr` on as well... which pure doesn't... I'm still not sure why the line is getting swallowed, but it doesn't appear to be because of this issue.

mafredri:
Hmm, I see value in this PR in following queue with zsh changes, but it will have no practical effect. But then again, I just updated to 5.4.1 (actually `HEAD-0f8cf76`) and Pure is behaving as expected, with or without the change.

screenshot 2017-08-10 23 52 09

Is this what you're seeing as well?

Even with default zsh options, no issues:

```console
~
❯ zsh -f
Franky% fpath+=(~/.zsh/functions)
Franky% autoload -Uz promptinit; promptinit
Franky% prompt pure

~
❯ echo -n hello world
hello world
~
❯
```

belak:
Hm, that's fun... I guess it means there's something in prezto triggering that. :/

belak:
I've narrowed it down to what line is causing the problems, so I'll close this. Sorry for the confusion!

https://github.com/sorin-ionescu/prezto/issues/1426

PR #349, opened 2017-09-29
More custom colors for the theme

vabatta:
Parametrised some colors of the theme for more customisation.  

`PURE_PREPROMPT_COLOR`  
Defines the color for the prepromt line. The default value is `blue`.  

`PURE_SUCCESS_COLOR`  
Defines the color for the prompt with exit code `0`. The default value is `magenta`.  

`PURE_ERROR_COLOR`  
Defines the color for the prompt with exit code different than `0`. The default value is `red`.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -324,6 +324,12 @@ prompt_pure_async_callback() {
 }
 
 prompt_pure_setup() {
+  # Define colors for BSD ls.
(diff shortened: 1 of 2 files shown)


mafredri:
Thanks for the PR, unfortunately we're not accepting PR's for colors that introduce new global variables for configuration. A PR or proposal for colors should use `zstyle` for configuration, please see #256 and #306 (also #292) where this is currently being discussed.

PR #350, opened 2017-09-29
Add the possibility to disable the virtual env prompt display.

classner:
This PR checks whether `VIRTUAL_ENV_DISABLE_PROMPT` is set before the setup is called and if so, honors it. This is implemented by setting the additional `PURE_VIRTUAL_ENV_DISABLE_PROMPT` variable, which disables the display.

This leads to a transparent and consistent behavior with what is expected if the user sets `VIRTUAL_ENV_DISABLE_PROMPT`. In my use case, I load a certain virtual env by default and do not want to have this very name always displayed.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -178,7 +178,9 @@ prompt_pure_precmd() {
 
 	# store name of virtualenv in psvar if activated
 	psvar[12]=
-	[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+	if [ -z "${PURE_VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
+		[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+	fi
 
 	# print the preprompt
 	prompt_pure_preprompt_render "precmd"
@@ -430,7 +432,13 @@ prompt_pure_setup() {
 	export PROMPT_EOL_MARK=''
 
 	# disallow python virtualenvs from updating the prompt
-	export VIRTUAL_ENV_DISABLE_PROMPT=1
+	if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ]; then
+		# it was not set. set it to only have pure modify the shell.
+		export VIRTUAL_ENV_DISABLE_PROMPT=1
+	else
+		# it was already set. also disable pure display.
+		export PURE_VIRTUAL_ENV_DISABLE_PROMPT=1
+	fi
 
 	prompt_opts=(subst percent)
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -176,10 +176,25 @@ prompt_pure_precmd() {
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks
 
-	# store name of virtualenv in psvar if activated
-	psvar[12]=
-	if [ -z "${PURE_VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
-		[[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+  # handling of virtualenv
+  # The following states of the environment variable
+  # $VIRTUAL_ENV_DISABLE_PROMPT are used:
+  # - undefined or 0: set it to 2,
+  # - set and >0: set it to 3,
+  # - 2: show virtual env,
+  # - 3: don't show virtual env.
+  # Thus, the variable space remains uncluttered.
+  psvar[12]=
+	if [[ ! -n "${VIRTUAL_ENV_DISABLE_PROMPT}"  ||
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -430,7 +432,13 @@ prompt_pure_setup() {
 	export PROMPT_EOL_MARK=''
 
 	# disallow python virtualenvs from updating the prompt
-	export VIRTUAL_ENV_DISABLE_PROMPT=1
+	if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ]; then
+		# it was not set. set it to only have pure modify the shell.
+		export VIRTUAL_ENV_DISABLE_PROMPT=1
+	else
+		# it was already set. also disable pure display.
+		export PURE_VIRTUAL_ENV_DISABLE_PROMPT=1
```
mafredri: I would like to avoid the extra global variable, how about we set `VIRTUAL_ENV_DISABLE_PROMPT=2` instead of introducing a new variable? It's a bit more hidden but should serve the same purpose.

classner:
Hi @mafredri!

Thanks a lot for the quick response and your detailed comments! I'll take all the code comments into account and make the requested changes. On the general concerns:

* "People might have previously set VIRTUAL_ENV_DISABLE_PROMPT and are now wondering why Pure is not displaying the virtualenv"
This is actually the expected bahavior that has been broken with the latest update. I did have VIRTUAL_ENV_DISABLE_PROMPT set since the last year and was using `pure` happily and was wondering why on earth since the latest update the variable is not 'working' any more. After several other things, I was looking through the pure code and found the reason: that it was not respecting it. I would strongly vote for respecting this variable: if someone wants to change the appearance of the shell they'll quickly find documentation for the variable in many sources. On the other hand, for a `pure` specific treatment, this may not be the case. In any case, it's your project and your decision.
* "Order matters, if someone sets VIRTUAL_ENV_DISABLE_PROMPT after prompt pure, Pure will continue to show the virtualenv". That's a very good point! I could remove the logic completely from `prompt_pure_setup` and have it in `prompt_pure_precmd`, then it would be independent of when someone sets the variable.

mafredri:
> This is actually the expected bahavior that has been broken with the latest update. I did have VIRTUAL_ENV_DISABLE_PROMPT set since the last year and was using pure happily and was wondering why on earth since the latest update the variable is not 'working' any more.

Sounds reasonable, I don't see any real downsides to honoring the user choice here and it seems like the more correct option.

> That's a very good point! I could remove the logic completely from prompt_pure_setup and have it in prompt_pure_precmd, then it would be independent of when someone sets the variable.

👍

classner:
Hi @mafredri,

I updated the PR as discussed. Let me know if you have questions or comments.

Christoph

pfrybar:
I updated pure last night, to discover that virtualenv was being displayed in my prompt even though I had `VIRTUAL_ENV_DISABLE_PROMPT=1`. I dug into the code this morning and came up with a quick fix (very similar to what classner had before):

```
diff --git a/pure.zsh b/pure.zsh
index cd55a3f..3ce4525 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -173,7 +173,7 @@ prompt_pure_precmd() {
 
        # store name of virtualenv in psvar if activated
        psvar[12]=
-       [[ -n $VIRTUAL_ENV ]] && psvar[12]="${VIRTUAL_ENV:t}"
+       [[ -n $VIRTUAL_ENV ]] && [[ $PURE_VIRTUAL_ENV_DISABLE_PROMPT -ne 1 ]] && psvar[12]="${VIRTUAL_ENV:t}"
 
        # print the preprompt
        prompt_pure_preprompt_render "precmd"
@@ -430,6 +430,9 @@ prompt_pure_setup() {
        # Prevent percentage showing up if output doesn't end with a newline.
        export PROMPT_EOL_MARK=''
 
+       # store virtualenv disable prompt in separate variable
+       PURE_VIRTUAL_ENV_DISABLE_PROMPT=$VIRTUAL_ENV_DISABLE_PROMPT
+
        # disallow python virtualenvs from updating the prompt
        export VIRTUAL_ENV_DISABLE_PROMPT=1
```

I think this approach is much simpler. It introduces a new global variable, but :man_shrugging: . About the ordering, I don't think this is a problem since you must already set VIRTUAL_ENV_DISABLE_PROMPT before sourcing virtualenv.

mafredri:
Hey @classner, first off, I want to apologize for not keeping up with this PR. There were a few things I wanted to address but did not find the time to think them through and explain properly. One was to reduce the number of states. The other, to avoid touching user-set values of `VIRTUAL_ENV_DISABLE_PROMPT`.

I believe I've addressed them in #381.

I'm still grateful for your PR and efforts in this matter!

@pfrybar I liked the simplicity of your solution, but it lacked the ability to react to runtime changes in `VIRTUAL_ENV_DISABLE_PROMPT`, still, thanks. You're implementation also contributed to #381.

If you either of you have the time, I would love it if you could verify that the behavior in #381 solves your issues.

PR #351, opened 2017-10-02
Add sp to the prompt_opts array

ericbn:
to fix `nopromptsp` being set in zsh 5.4, which hides the last line of outputs without a LF character at the end, like

    echo -n "bla"

With zsh starting at version 5.4, when using the `prompt_opts` array, `nopromptsp` will be set unless the array contains a `sp`. This was introduced in zsh at https://github.com/zsh-users/zsh/commit/43e55a9bcd2c90124a751f2597d2f33cb6e3c042#diff-bb10d67e7a8561b66a53a805f3c77a40R233

See https://github.com/Eriner/zim/issues/209

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -432,11 +432,11 @@ prompt_pure_setup() {
 	# disallow python virtualenvs from updating the prompt
 	export VIRTUAL_ENV_DISABLE_PROMPT=1
 
-	prompt_opts=(subst percent)
+	prompt_opts=(percent sp subst)
 
 	# borrowed from promptinit, sets the prompt options in case pure was not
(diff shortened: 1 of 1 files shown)


mafredri:
This change will have no effect, `prompt_sp` is a no-op without `prompt_cr`. With both `sp` and `cr` set, the prompt produces unwanted whitespace at the end of all output. Pure already prints it's own newline at each prompt which makes this behavior less useful.

Arguably, enabling these options would be more correct in terms of output, but IMO it's not worth the tradeoffs (whitespace):

screenshot 2017-10-02 16 52 18


Thanks for the consideration, though!

PR #353, opened 2017-10-04
Reproposal of colors' theme options with zstyle

vabatta:
As in https://github.com/sindresorhus/pure/pull/256, this is a proposal for those theme customisations using `zstyle` to set them avoiding creating new globals.

### DIFF
--- a/async.zsh
+++ b/async.zsh
@@ -3,20 +3,20 @@
 #
 # zsh-async
 #
-# version: 1.5.0
+# version: 1.5.2
 # author: Mathias Fredriksson
 # url: https://github.com/mafredri/zsh-async
 #
 
 # Produce debug output from zsh-async when set to 1.
-ASYNC_DEBUG=${ASYNC_DEBUG:-0}
+typeset -g ASYNC_DEBUG=${ASYNC_DEBUG:-0}
 
 # Wrapper for jobs executed by the async worker, gives output in parseable format with execution time
 _async_job() {
 	# Disable xtrace as it would mangle the output.
 	setopt localoptions noxtrace
 
-	# Store start time as double precision (+E disables scientific notation)
+	# Store start time for job.
 	float -F duration=$EPOCHREALTIME
 
 	# Run the command and capture both stdout (`eval`) and stderr (`cat`) in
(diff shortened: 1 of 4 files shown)


mcallan83:
I would love to see this merged in. It seems like a lot of people have asked. Pure is great and I have used it for years now, I just wish I had a but more control over the colors. The contrast in the current colors is not enough for me to see without straining. Also, since everyone uses a different color scheme, this would help in that area as well. Would 2 or 3 extra variables really hurt things?

mafredri:
Thanks for the PR @vabatta and sorry for leaving this hanging for so long.

Could you rebase this on master and remove all non-color related changes so that we don't introduce any backwards incompatibilities? As it stand this PR would break a lot of user settings.

vabatta:
@mafredri I've done the rebase, but create a branch for the transition so you can test it out before merging it to the master.

mafredri:
Thanks @vabatta. This PR still contains changes unrelated to colors though. I want this PR only to focus on colors and not touch any other settings.

That keeps this PR small and backwards compatible. And it will allow us to see how the whole zstyle pattern fits before we go all in. 

If you want you can open up a new PR for the other changes so we can discuss them there.

vabatta:
@mafredri You mean the changes on `readme.md` or in which other file?

mafredri:
Both readme and pure.zsh, e.g. `zstyle -a ":pure:prompt" cmd_max_exec_time option_cmd_max_exec_time`, `zstyle -a ":pure:git" pull git_pull`, etc. Those aren't color related and are breaking changes.

vabatta:
@mafredri you're right, I actually changed every single variable using `zstyle` and I wasn't even realizing that. I suggest you to open up a new branch where I can PR without mess up the `master` (for example `color-development`).

mafredri:
@vabatta no need to worry about messing up master. Feel free to push to this PR branch (vabatta:master) and I'll be able to review and test it 😄.

vabatta:
@mafredri I need to create a new PR with only the changes you need.

vabatta:
@mafredri here you go, only the changes you need for colours.

eyal-lezmy:
Do you think it is possible to also configure the [`git_branch`](https://github.com/sindresorhus/pure/pull/353/files#diff-ca464c5c9e47e8a26c7e6ac0ed786406L105) color from zstyle?

PR #368, opened 2017-11-20
Prevent multiple prompt resets in one execution cycle

### LINKED ISSUE #356: Sometimes Pure clears the output of the previous line
### General information

- Pure version: 1.5.2
- ZSH version: 5.4.2
- Terminal program & version: iTerm 3
- Operating system: macOS 10.13
- ZSH framework: None

I have:
- [x] Tested with another terminal program and can reproduce the issue: 

### Problem description

```
~/dev/oss/npm-home master*

~/dev/oss/npm-home master ⇡
❯ o

~/dev/oss/npm-home master ⇡
❯ o
```

Notice how `o` is missing from the first prompt line. I wrote and executed it there, but it got cleared right after. I also see this problem pretty often with the last line being cleared of the output of [`np`](https://github.com/sindresorhus/np), [relevant issue](https://github.com/sindresorhus/np/issues/158).

It seems like it's being cleared when `⇡` is added.

This issue only happens once in a while and not consistently, which makes me believe it's something with the async setup and maybe a race issue.

mafredri:
The prompt ends up in a weird state when `zle reset-prompt` is called multiple times during one execution cycle. This can happen when multiple async tasks finish at nearly the same time. What happened here is that `async_process_results` called `prompt_pure_async_callback` multiple
times during it's execution. In turn, `prompt_pure_async_callback` did multiple calls to `zle reset-prompt`. My theory is that ZLE ends up in a weird state that is reset after all the current code has completed executing.

This behavior is observable when the prompt "moves upwards" and erases previous lines in the terminal.

Fixes #356.

### DIFF
--- a/async.zsh
+++ b/async.zsh
@@ -3,7 +3,7 @@
 #
 # zsh-async
 #
-# version: 1.5.2
+# version: 1.6.0
 # author: Mathias Fredriksson
 # url: https://github.com/mafredri/zsh-async
 #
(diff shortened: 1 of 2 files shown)


sindresorhus:
Yay. Super happy to have this fixed ✨

PR #378, opened 2018-02-15
Add optional prompt prefix if root

pmccarren:
I propose we add an optional prefix to the prompt if we are root. I find the extra visibility to be of value when working on remote boxes in production.

The proposed prefix is adding a red `#`.
e.g. `# ❯`

screenshot

We should default to hiding the prefix unless configured to do otherwise. This is accomplished by checking for the existence of the PURE_ROOT_HASH variable, which is by default not set.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -464,6 +464,11 @@ prompt_pure_setup() {
 	# if a virtualenv is activated, display it in grey
 	PROMPT='%(12V.%F{242}%12v%f .)'
 
+	# if PURE_ROOT_HASH is set and we are root then prefix prompt with a red hash '#'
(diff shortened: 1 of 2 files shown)


mafredri:
I don't think this is a common enough use case, and it wouldn't feel right to introduce yet another option just to have multiple indications of the same thing (there's already the white `root`).

You can easily achieve the same thing by setting the prompt symbol in your `.zshrc`:

```zsh
PURE_PROMPT_SYMBOL='%F{red}#%f ❯'
```

Does that work for you?

mafredri:
This has been [added to the readme](https://github.com/sindresorhus/pure/wiki/Home/c4693299be6a67674838dcc415c86104b55f8f85#prefix-prompt-when-logged-in-as-root). Even though this didn't turn out to be a good fit for Pure, we appreciate your contribution :).

PR #380, opened 2018-02-24
Add variable for overriding default prompt symbol color

arunvelsriram:
I have added a new variable `PURE_PROMPT_SYMBOL_COLOR` to override the default color `magenta`.

Since it is hard to differentiate between `magenta` and `red` there is a possibility that we might fail to notice when a command exits with a non-zero status.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -465,7 +465,7 @@ prompt_pure_setup() {
 	PROMPT='%(12V.%F{242}%12v%f .)'
 
 	# prompt turns red if the previous command didn't exit with 0
-	PROMPT+='%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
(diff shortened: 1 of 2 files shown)


mafredri:
Closing this in favor of #353.

By the way, you can easily achieve this by setting the prompt symbol:

```zsh
PURE_PROMPT_SYMBOL='%F{red}❯%f'
```

arunvelsriram:
Waiting for the referenced PR to be merged.

`PURE_PROMPT_SYMBOL='%F{red}❯%f'` will always keep the prompt symbol in the specified color. It will not change to red when a command returns with non-zero exit status.

mafredri:
Ah, good point, at least this would work:

```zsh
PURE_PROMPT_SYMBOL='%f%(?.%F{green}.%F{red})❯'
```

PR #383, opened 2018-03-10
Add env var DISABLE_AUTO_TITLE

veggiemonk:
Using iterm2 on Mac.
This PR allows the user to set the title of the window/tab

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -56,6 +56,11 @@ prompt_pure_set_title() {
 	# emacs terminal does not support settings the title
 	(( ${+EMACS} )) && return
 
+	# disable auto title 
+	if "$DISABLE_AUTO_TITLE" == true; then
+		return
+	fi
+
 	# tell the terminal we are setting the title
 	print -n '\e]0;'
 	# show hostname if connected through ssh


pure.zsh
```diff
@@ -56,6 +56,11 @@ prompt_pure_set_title() {
 	# emacs terminal does not support settings the title
 	(( ${+EMACS} )) && return
 
+	# respect oh-my-zsh disable auto title 
+	if [[ "$DISABLE_AUTO_TITLE" == true ]]; then
```
mafredri: Avoid quoting the variable inside `[[ ]]`, it's not required.

veggiemonk:
Hi @mafredri,

You're totally right about the OMZ support. I don't think it make sense to support every framework out there.

On the other hand it took me quite a while to figure out why the title wouldn't change even if I set it manually. Hence this PR. I thought it made total sense to have that option in Pure. And it doesn't hurt that it has the same name as the OMZ framework. 😄 

My use case is that with new macbook, the touch bar displays all the tab of terminal. it allows to change tab as quick as a keystroke (touchstroke ?). But if the title is a long PATH, it is basically useless because that's the only thing on the touch bar.

I'll make the change to the PR. Let me know what you think and no hard feelings if this don't get merged. I've used Pure for a long time and it just made sense to contribute back.

pure.zsh
```diff
@@ -56,6 +56,11 @@ prompt_pure_set_title() {
 	# emacs terminal does not support settings the title
 	(( ${+EMACS} )) && return
 
+	# disable auto title 
+	if "$DISABLE_AUTO_TITLE" == true; then
```
mafredri: This still requires the `[[ ]]` around it.

mafredri:
@veggiemonk sorry for the delay! At the very least, for this to make sense in Pure, it *also* needs to accept `1` as a value (to match other Pure options). And it needs to be added to the readme.

Btw, there's another way to bypass setting the title:

```
# First, load Pure.
prompt pure

# Then, replace the set title function with an empty one.
prompt_pure_set_title() {}
```

---

@sindresorhus do you have any preference on this?

veggiemonk:
@mafredri Well, the solution you gave is much better than this PR. It should be added to the README though because I spent some time looking for that and I didn't even think about it.

Since most env var starts with `PURE`, having one that does not break the consistency.

mafredri:
I've added this to the Wiki: [Disable Pure terminal title updates](https://github.com/sindresorhus/pure/wiki#disable-pure-terminal-title-updates). I don't want to add it to the readme since that would sanction such practices and they could be considered *supported*. The user should be aware that this can break at any moment and we don't officially support it 😄.

PR #393, opened 2018-04-25
Try to detech SSH connection when SSH_CONNECTION is unset

### LINKED ISSUE #382: Hostname not showing in the window title after sudo su -
### General information

- Pure version: 1.6.0
- ZSH version: 5.0.2
- Terminal program & version: PuTTY 0.70 & KiTTY 0.70
- Operating system: Windows 10
- ZSH framework: oh-my-zsh

### Problem description

The window's title doesn't show the hostname when sudoing from a user.

### Reproduction steps

1. Connect to a server with a regular user
2. sudo su -
3. The hostname is not showing anymore in the window title

With a regular user :
image

After sudo su - :
image

### My `.zshrc`:

```shell
prompt pure
```

mafredri:
This change allows us to detect SSH connections even when the `$SSH_CONNECTION` environment variable is gone. This can happen when a user changes to another user (`su`, `sudo`) or the environment is otherwise reset.

By checking the output of `who am i` for a parenthesis at the end, we assume that said parenthesis must contain either an IPv4 or IPv6 address, indicating that this is a remote session.

The reliablity of this assumption has not been analyzed to a great degree, there could be false positives, or systems that do not format the output in this way.

Fixes #382.

This PR might introduce a start-up impact (via `who am i | grep -q '(.*)$'`) but I'm not really worried that it would be noticeable.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -134,7 +134,7 @@ prompt_pure_preprompt_render() {
 	fi
 
 	# Username and machine, if applicable.
-	[[ -n $prompt_pure_username ]] && preprompt_parts+=('$prompt_pure_username')
+	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=('${prompt_pure_state[username]}')
 	# Execution time.
 	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{yellow}${prompt_pure_cmd_exec_time}%f')
 
@@ -445,6 +445,8 @@ prompt_pure_async_callback() {
 }
 
 prompt_pure_setup() {
+	setopt localoptions noshwordsplit
+
 	# Prevent percentage showing up if output doesn't end with a newline.
 	export PROMPT_EOL_MARK=''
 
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -470,11 +472,25 @@ prompt_pure_setup() {
...
-	[[ "$SSH_CONNECTION" != '' ]] && prompt_pure_username='%F{242}%n@%m%f'
+	[[ -n $ssh_connection ]] && username='%F{242}%n@%m%f'
 
 	# show username@host if root, with username in white
-	[[ $UID -eq 0 ]] && prompt_pure_username='%F{white}%n%f%F{242}@%m%f'
+	[[ $UID -eq 0 ]] && username='%F{white}%n%f%F{242}@%m%f'
+
+	typeset -gA prompt_pure_state=(
+		username "$username"
+	)
```
mafredri: I'm considering doing more checks at "startup" in the future, the result of those would also go into `prompt_pure_state`. We should be able to get rid of a bunch of other global (internal) pure variables this way.

sindresorhus: I like it 👍

mafredri:
Thanks for reviewing! I'm fairly confident it will work based on my tests. There might be some odd implementations of `who` out there where it might not work, and I'm not sure if there can be false positives (e.g. are there circumstances where `who` includes parens at the end but it's not over an SSH connection).

As I was writing this, I decided to take a look at `coreutils`. Seems parenthesis might be added in X sessions as well: [coreutils/src/who.c#L374-L432](https://github.com/coreutils/coreutils/blob/22424dde74d0b3a80c6f046c1e0194c580fcb7e2/src/who.c#L374-L432). Going to test in a VM.

mafredri:
Turned out my hunch was right, #398 fixes it.

PR #396, opened 2018-04-27
Improve the debug prompt (PS4)

mafredri:
I've long wanted to do something about the debug prompt (PS4) to improve usability / visibility. By default, it's not very easy to do a quick scan of the output.

This PR essentially does three things:
- Add colors, to highlight the current function and de-empasize line numbers
- Add an extra line number (`%I`) which represents the line number in the file where the code resides (by default, only the function line number is shown, making it hard(er) to find the relevant code)
- Repeat the + character to indicate the current execution depth, e.g. when a function calls another, the depth is increased
    - This allows you to quickly see which function invoked which

This is the result (when launched via `zsh -x` or `setopt XTRACE`):

screen shot 2018-04-27 at 20 40 14

Original looks like this:

screen shot 2018-04-27 at 19 40 27

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -533,13 +533,10 @@ prompt_pure_setup() {
 	# prompt turns red if the previous command didn't exit with 0
 	PROMPT+='%(?.%F{magenta}.%F{red})${PURE_PROMPT_SYMBOL:-❯}%f '
 
-	# Store %e (execution depth) for in-place expansion via (S%).
+	# Store prompt expansion symbols for in-place expansion via (%). For
+	# some reason it does not work without storing them in a variable first.
 	typeset -ga prompt_pure_debug_depth
 	prompt_pure_debug_depth=('%e' '%N' '%x')
-
-	# Improve the debug prompt (PS4) with colors to highlight essential
-	# parts, show depth by repeating the +-sign and include the line number
-	# where the code resides (%I).
 
 	# Compare is used to check if %N equals %x. When they differ, the main
 	# prompt is used to allow displaying both file name and function. When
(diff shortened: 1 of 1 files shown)


mafredri:
I would love to include the file path (`%x`), but including both fn name and file path is kinda crowded. Which is more important / better? I don't know. With file path, it might look something like this:

screen shot 2018-04-27 at 20 21 03

sindresorhus:
Looks good

sindresorhus:
> I would love to include the file path (%x), but including both fn name and file path is kinda crowded.

You could include only the filename, but let it be a link to the file: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname

mafredri:
I tried the links with iTerm but the performance impact was pretty significant 😞. Haven't tried any other terminals but I can imagine that (at least for now) this feature would hinder debugging rather than enable it, sadness.

mafredri:
I dove a bit deeper, figured there had to be a way to bend zsh to my will...

#### Add filename to debug prompt (1ff12ff)

This commit also adds the filename (without path) to the debug prompt, when applicable. It also makes sure to avoid displaying duplicate information, say, when a file is being parsed, both %N and %x contain the same value (file path).

There is a slight cost to doing the extra processing when testing in iTerm (unscientific benchmark):

	zsh -x  0.30s user 0.14s system 57% cpu 0.754 total

vs

	zsh -x  0.49s user 0.16s system 65% cpu 0.988 total

I think this is within an acceptable range.

---

This is the new look (I used `@` here btw to hint at the function being a symbol, as in many editors):

screen shot 2018-05-06 at 17 56 31

And this is what it looks like when the alternate debug prompt kicks in:

screen shot 2018-05-06 at 18 31 35

So when there is no function being executed, the full path is shown (because we have extra space).

*NOTE:* `add-zsh-hook` provides duplicate info here but it's actually behaving as it should (`%N` is `add-zsh-hook` and `%x` is `.../share/zsh/functions/add-zsh-hook`). I could technically get rid of it by further comparing the output between `%N` and `${${(%)prompt_pure_debug_depth[3]}:t}`, but that would make it even more complicated. I opted not to.

salmanulfarzy:
This is awesome. Thank you @mafredri :smile:

PR #398, opened 2018-04-29
Prevent hostname from showing up in local X sessions

mafredri:
This change prevents the hostname from showing up when a terminal is opened inside a local X session (displayed as (:1)).

Fixes a regression introduced in #393.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -479,7 +479,8 @@ prompt_pure_async_callback() {
 prompt_pure_state_setup() {
 	setopt localoptions noshwordsplit
 
-	local ssh_connection=$SSH_CONNECTION
+	# Check SSH_CONNECTION and the current state.
+	local ssh_connection=${SSH_CONNECTION:-$prompt_pure_detected_ssh_connection}
 	local username
 	if [[ -z $ssh_connection ]] && (( $+commands[who] )); then
 		# When changing user on a remote system, the $SSH_CONNECTION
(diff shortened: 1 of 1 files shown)


open-dynaMIX:
I was just writing an issue for that regression, then I saw this PR. I'd like to add, that on my system (up-to-date Arch Linux with `who (GNU coreutils) 8.29`) I get the following:
 ```
❯ who am i
myusername    pts/1        2018-04-30 19:25 (f)
```
This will not be handeled by `grep -v '(:[0-9]*)$'`.

mafredri:
Interesting, thanks for reporting, what does `f` mean?

open-dynaMIX:
Unfortunately I don't know. The only thing that comes to mind is, that it's the first character of my hostname and username.

open-dynaMIX:
Just realized: This happens in guake and in xterm. In termite `who am i` does not even output anything. But I guess that could be an issue unrelated to pure...

open-dynaMIX:
On my Arch system with `who (GNU coreutils) 8.29` `--ips` is not recognized, but on one of my debian VMs with `who (GNU coreutils) 8.26` `--ips` it is...

mafredri:
Ok, so I went back to the drawing board with this one. @open-dynaMIX would you be willing to try this out?

I rebased this on the latest master, but the changes here are essentially:

#### Implement a more robust remote connection detection (who) (1283d83)

First, we switch from `who am i` to `who -m` because this is the POSIX way and macOS supports it as well.

Second, if the above fails, we fall back to using just `who` and grep the line that represents our TTY. This works e.g. with busybox systems.

Finally, we use (extremely) simplified regular expressions for detecting IPv4, IPv6 and hostname. This is a best-effort attempt at detecting the remote connections when SSH_CONNECTION is not set.

#### Inherit SSH connection detection inside tmux or other subshells (c801cfa)

open-dynaMIX:
Nice! Unfortunately just yesterday I killed my last remote host, where I had zsh and pure installed, so I was only able to try it with SSH into localhost (from `::1` and from `127.0.0.1`). It detects the SSH session correctly in both cases.

Thank you for your work on this!

mafredri:
Thanks for testing @open-dynaMIX! This has now landed on `master`.

PR #407, opened 2018-05-07
Show options as a table in readme

mafredri:
Been meaning to do this for a while, it should make the options section a bit more compact and easier to scan.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -81,33 +81,15 @@ prompt pure
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
As a maintainer, I dislike Markdown tables as they're super annoying to edit, but as a user, I can see how this looks better.

mafredri:
I hear 'ya! I had help from disabling line wrapping and having an automatic table formatting extension for my editor, otherwise I would have died.

PR #413, opened 2018-06-17
Prevent IPv6 regexp from capturing the time

### LINKED ISSUE #403: False SSH connection detected
### General information

- Pure version: up to date master
- ZSH version: 5.5.1
- Terminal program & version: rxvt-unicode v9.22
- Operating system: Debian Stretch
- ZSH framework: none

### Problem description

This [commit](https://github.com/sindresorhus/pure/commit/17e23aecda563f6135b91d0d2eea82dadb0ca070) causes the appearance of `user@host` in my prompt even when I'm not connected via ssh. Some investigation:

```
echo "@${SSH_CONNECTION}@" -> @@
echo $+commands[who]       -> 1
who am i                   -> user pts/6 2018-05-04 19:10 (:0)
```

mafredri:
Fixes #403.

Test input used (+ = match, - = no match):

```diff
- :1
- 1:
- a:1
- 1:a
- 10:22
- ::
+ ::1
- 1::
+ fe::1
+ 2001:0000:3238:DFE1:0063:0000:0000:FEFB
+ 2001:0000:3238:DFE1:63:0000:0000:FEFB
+ 2001:0000:3238:DFE1:63::FEFB
+ 2001:0:3238:DFE1:63::FEFB
+ 2001::FEFB
```

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -497,7 +497,7 @@ prompt_pure_state_setup() {
 			who_out=$(who 2>/dev/null | grep ${TTY#/dev/})
 		fi
 
-		local reIPv6='([a-f0-9:]+:+)+[a-f0-9]+'  # Simplified, but matches loopback as well (::1).
+		local reIPv6='(([0-9a-fA-F]+:)|:){2,}[0-9a-fA-F]+'  # Simplified, only checks partial pattern.
 		local reIPv4='([0-9]{1,3}\.){3}[0-9]+'   # Simplified, allows invalid ranges.
 		# Here we assume two non-consecutive periods represents a
 		# hostname. This matches foo.bar.baz, but not foo.bar.


jenniges:
Works for me. Tested also with a variaty of nested tmux and ssh connections to localhost.
I'm not a network expert, but as far as I know, `::` and `1::` are valid ipv6 addresses. If not, just ignore me.
The regex I mentioned in #403 matches them as well. :-)

 `([0-9a-fA-F]*:){2,}[0-9a-fA-F:]*`
```diff
- :1
- 1:
- a:1
- 1:a
- 10:22
+ ::
+ ::1
+ 1::
+ fe::1
+ 2001:0000:3238:DFE1:0063:0000:0000:FEFB
+ 2001:0000:3238:DFE1:63:0000:0000:FEFB
+ 2001:0000:3238:DFE1:63::FEFB
+ 2001:0:3238:DFE1:63::FEFB
+ 2001::FEFB
```

mafredri:
Thanks for testing @jenniges. And you do have a valid point about those addresses being valid. Although, AFAICT, `::` can't be configured as an address (although it can be seen in some? places). And while `X::` can be used, I think recommended to avoid them.

TBH, I'm very unfamiliar with IPv6 and don't know what we reasonably can expect. But in this case, I have a feeling we should avoid ambiguity as much as possible.

With the regexp you created, we would also match things like `1:1:` and `:1:`. Not sure where we should draw the line with keeping the regexp simple, correct and unambiguous.

jenniges:
You are right. Your solution solves the initial problem and we will probably never come in touch with an IP like `::` or `X::`, so better stay with yours.

mafredri:
I appreciate your input @jenniges, let's hope that's the case 🤞. I'm merging this now so we have a fix on master, I'll revisit your solution should there be any problems with this approach 😄.

0xC0FFEE:
This fixes #403 for me too. Thanks!

PR #418, opened 2018-07-05
prompt_pure_string_length_to_var() is not used

dennisse:
This function has not been used since commit https://github.com/sindresorhus/pure/commit/4cdd0cf4f48d0c2064700c4fc60083bc0b3f6693, when we started performing git-checks asynchronously.


mafredri:
Good spot, thanks!

Issue #384, opened 2018-03-11
Non-npm install could be simplified


mattrixman:
Hello.  I think pure is awesome, I want to use it everywhere.  

The README made me think that installing pure on a server without `npm` or root access would be too cumbersome to do several times a day.  After looking closer, I see that it's actually quite easy to automate--but it would be even easier if small changes were made to this repo. 

Here is the install process as documented
```
1. pull pure from github
2. examine fpath
3. add a symlink in fpath to pure.zsh
4. add a symlink in fpath to async.zsh
5. modify zshrc:
        autoload -U promptinit; promptinit
        prompt pure
```

But this works too:
```
1. pull pure from github into <pure_dir>
2. copy `pure.zsh` to `prompt_pure_setup`
3. copy `async.zsh` to `async`
4. modify .zshrc:
        fpath+=("<pure_dir>")
        autoload -U promptinit; promptinit
        prompt pure
```

Steps 2 and 3 could be eliminated by either:
- rename the files in the repo so that the user doesn't have to do it by hand
- modify whatever is looking for `prompt_pure_setup` so that it looks for `pure.zsh` too (and likewise for async)

So the resulting install would just be:
```
1. pull pure from github into <pure_dir>
2. modify .zshrc:

        fpath+=("<pure_dir>")
        autoload -U promptinit; promptinit
        prompt pure
```

The modifications above would also be helpful for cases where the user uses git to sync parts their home directory between servers (like I do).  As it stands, [my zshrc](https://github.com/MatrixManAtYrService/home/blob/master/.zshrc) contains this monster:

```
# pure prompt (https://github.com/sindresorhus/pure)
PROMPT_DIR="$HOME/.zsh/pure"
PROMPT_SETUP_IN="$HOME/.zsh/pure/pure.zsh"
PROMPT_SETUP="$HOME/.zsh/pure/prompt_pure_setup"
PROMPT_ASYNC_IN="$HOME/.zsh/pure/async.zsh"
PROMPT_ASYNC="$HOME/.zsh/pure/async"

if [ -d "$PROMPT_DIR" ] ; then
    # If the pure submodule has been fetched
    # align the setup files with pure's expectations
    if [ ! -f "$PROMPT_SETUP" ] ; then
        cp -v "$PROMPT_SETUP_IN" "$PROMPT_SETUP"
    fi
    if [ ! -f "$PROMPT_ASYNC" ] ; then
[...]

PR #419, opened 2018-07-16
Include symlinks in the repo to simplify setup process

MatrixManAtYrService:
See discussion here: https://github.com/sindresorhus/pure/issues/384

This puts the symlinks `async` and `prompt_pure_setup` right in the repo, which removes the necessity of creating them during install.  It is less transparent, because the user doesn't learn which two files need to be in `fpath` but I think the increase in  simplicity is worth it.

### DIFF
--- a/prompt_pure_setup
+++ b/prompt_pure_setup
@@ -1,0 +1,1 @@
+pure.zsh
--- a/readme.md
+++ b/readme.md
@@ -34,39 +34,15 @@ $ npm install --global pure-prompt
 
 That's it. Skip to [Getting started](#getting-started).
 
-### Manually
+### Manually (per user)
 
-1. Either…
-  - Clone this repo
-  - add it as a submodule, or
-  - just download [`pure.zsh`](pure.zsh) and [`async.zsh`](async.zsh)
+1. Clone this repo into /some/dir
 
-2. Symlink `pure.zsh` to somewhere in [`$fpath`](https://www.refining-linux.org/archives/46-ZSH-Gem-12-Autoloading-functions.html) with the name `prompt_pure_setup`.
+2. modify .zshrc to include
 
-3. Symlink `async.zsh` in `$fpath` with the name `async`.
-
-#### Example
-
-```console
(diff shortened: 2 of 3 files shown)


prompt_pure_setup
```diff
@@ -0,0 +1 @@
+pure.zsh
```
mafredri: 1. How about moving these into a folder called `fpath` to keep the project directory cleaner?
2. Could you turn these into sourced files instead of symlinks? See [here](https://github.com/mafredri/zsh-async/blob/5248bd85bf2c522ebfa2b8d18ee7bc736d10c791/async.plugin.zsh) for an example (https://github.com/mafredri/zsh-async/pull/22#issuecomment-384260603) for discussion.

Regarding 2, symlinks can be disabled in git and as such, won't work for everyone.

readme.md
```diff
@@ -36,38 +36,14 @@ That's it. Skip to [Getting started](#getting-started).
...
 ### Manually
 
-1. Either…
-  - Clone this repo
-  - add it as a submodule, or
-  - just download [`pure.zsh`](pure.zsh) and [`async.zsh`](async.zsh)
+1. Clone this repo into /some/dir
 
-2. Symlink `pure.zsh` to somewhere in [`$fpath`](https://www.refining-linux.org/archives/46-ZSH-Gem-12-Autoloading-functions.html) with the name `prompt_pure_setup`.
+2. modify .zshrc to include
```
mafredri: modify => Modify, end with :

readme.md
```diff
@@ -36,38 +36,14 @@ That's it. Skip to [Getting started](#getting-started).
 
 ### Manually
 
-1. Either…
-  - Clone this repo
-  - add it as a submodule, or
-  - just download [`pure.zsh`](pure.zsh) and [`async.zsh`](async.zsh)
+1. Clone this repo into /some/dir
```
mafredri: Maybe we could use a more descriptive folder name, like `$HOME/pure`? It might be nicer if the path is usable *out of the box*, so to speak.

sindresorhus:
@MatrixManAtYrService Ping :)

PR #430, opened 2018-08-20
Simplify async tasks by not passing $PWD

mafredri:
With zsh-async 1.7.0 we can update the environment of the async worker
(change directory) instead of always passing it as an argument.

This has the added benefit of keeping the async worker pwd in sync with
the terminal and will thus show the correct path when inspecting child
processes.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -236,7 +236,7 @@ prompt_pure_async_vcs_info() {
 	vcs_info
 
 	local -A info
-	info[pwd]=$1
+	info[pwd]=$PWD
 	info[top]=$vcs_info_msg_1_
 	info[branch]=$vcs_info_msg_0_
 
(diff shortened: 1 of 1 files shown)


mafredri:
This PR also fixes two bugs, where `$working_tree` (a non-existent variable) was used.

sindresorhus:
Can you rebase from master?

mafredri:
That was quick 😄, done 👍.

mafredri:
Looks like I haven't managed to fully fix https://github.com/sindresorhus/pure/pull/428. Going to be investigating that separate from this PR.

sindresorhus:
With this, I no longer see branch names.

mafredri:
@sindresorhus did your test include 
https://github.com/sindresorhus/pure/pull/430/commits/d54926840cfe1c753b6b76545125eebaddfdbe46? Branch names were broken before it, but it's been working fine for me now.

sindresorhus:
Yeah, that was it. Works great now.

PR #434, opened 2018-09-08
Do not run zle .reset-prompt when switching vi-mode if keys are queued

FrostedMint:
Hello!

I noticed an issue when using Pure with vi-mode. I haven’t tried this on an absolutely clean Zsh installation, but hopefully this repro works for you:

1. Switch to vi-mode with `$ setopt vi`
2. Run some command, like `$ print bark bark`
3. Enter some text in the buffer
4. As quickly as possible, hit `Esc` followed by `c` `c` (which is the vi command to clear the entire line and enter insert mode)

On my machine, this causes the prompt to be drawn one line above where it should be. This can be repeated to progressively move the prompt to the top of the screen, one line at a time.

While poking around trying to figure out what was happening, I noticed that `prompt_pure_update_vim_prompt_widget` was being invoked twice, once for the mode switch to `vicmd` mode, and again for the mode switch back to `main` (insert) mode. Furthermore, on the first invocation, `$KEYS_QUEUED_COUNT`, which is normally 0, was set to 1. For some reason, if this value is greater than 0, calling `zle .reset-prompt` will cause the prompt to be redrawn in the incorrect position.

My theory is that certain vim actions can be “queued up” in ZLE. When typing `Esc`, ZLE will wait a few moments before triggering the mode switch to `vicmd`. However, if we immediately follow by typing `c` `c`, since the first mode switch has yet to occur, we appear to “queue up” a second mode switch back to mode `main`. ZLE will then execute the `prompt_pure_update_vim_prompt_widget` hook twice in quick succession.

**Solution:**
In `prompt_pure_update_vim_prompt_widget`, don’t call `zle .reset-prompt` if `$KEYS_QUEUED_COUNT` is greater than 0. I’m not sure if this check is also needed in `prompt_pure_reset_vim_prompt_widget`, but my gut feeling is no.

**Background info:**
Running `zsh 5.5.1 (x86_64-apple-darwin17.5.0)` on macOS 10.13.6 (17G65). Issue occurs inside and outside of `tmux` (version 2.7), and in both iTerm2 (Build 3.1.7) and Terminal.app (version 2.8.2 (404)). Both `zsh` and `tmux` are binary builds (bottles) installed via Homebrew.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -484,7 +484,9 @@ prompt_pure_reset_prompt_symbol() {
 prompt_pure_update_vim_prompt_widget() {
 	setopt localoptions noshwordsplit
 	prompt_pure_state[prompt]=${${KEYMAP/vicmd/${PURE_PROMPT_VICMD_SYMBOL:-❮}}/(main|viins)/${PURE_PROMPT_SYMBOL:-❯}}
-	zle && zle .reset-prompt
+	if (( $KEYS_QUEUED_COUNT == 0 )) ; then
+		zle && zle .reset-prompt
+	fi
 }
 
 prompt_pure_reset_vim_prompt_widget() {


nbari:
Sharing a [gif](https://giphy.com/apps/giphycapture) would help to get a better idea, I have been using the vim feature and works fine so far, even with your example `esc + c c` only issue that I found is that prompt changes slow (double the time) when entering into insert mode but going back to normal mode  is always faster.

nbari:
@mafredri I just notice that the delay only happens within tmux any ideas?

FrostedMint:
Okay, tried in a minimal Zsh setup. Here are links to the [`pure.zsh`](https://raw.githubusercontent.com/sindresorhus/pure/9325fe60f8857ae089146176c2fa73384507dfc3/pure.zsh) and [`async.zsh`](https://raw.githubusercontent.com/sindresorhus/pure/16b141b11ab908bf21d07d21df536e94078a6f01/async.zsh) I’ve downloaded. And my `.zshrc`:

```zsh
set -o vi
fpath=( "$HOME/zsh-tmp/fndir" $fpath )
prompt pure
```

Here’s a gif with the default (I assume?) `$KEYTIMEOUT` of 40. This is outside of tmux:

kt40

If I just type `Esc`, there is a delay before the arrow indicator changes. If I hit `Esc` quickly followed by `b` (which keeps me in `vicmd` mode), the arrow indicator changes as soon as I hit `b`, but the prompt does not move.

However, when I quickly alternate between pressing `Esc` and `i`, the prompt moves. Notice that the indicator never has time to change to the `vicmd` mode indicator, because we switch back to insert before the key timeout.

FrostedMint:
Here’s a gif with `$KEYTIMEOUT` set to 1. In this case everything works. Notice that the indicator now changes when hitting `Esc` and `i`:

kt1

mafredri:
@FrostedMint thanks, I can reproduce with:

```console
> zsh -f
% bindkey -v
% KEYTIMEOUT=40
% fpath=(~/.zsh/functions $fpath)
% autoload -U promptinit; promptinit
% prompt pure
```

and just doing `Esc + i`. But interestingly, I can't reproduce with my own zsh configuration.

It also looks like `KEYS_QUEUED_COUNT` was added in Zsh 5.4, so this would not work (and produce an error) in older versions of Zsh. And it seems like other keys can be queued up as well, say `Esc + p` would leave the the prompt in `vicmd` mode, but the prompt would not be updated to reflect this. So it looks like we need to figure out another approach to tackle this 😕.

mafredri:
@FrostedMint are you still working on this?

FrostedMint:
@mafredri, my apologies, I am not, wasn’t sure how to proceed given the issues you brought up, and then I got distracted by life. 🍃

FrostedMint:
Closing for now since I haven’t touched this in months.

PR #440, opened 2018-10-16
Add conda environment name to precmd

sbmueller:
I've seen this issue discussed in the issues, but it has not found it's way into the code.
Credit goes to @minhoolee and @VincentMatthys .
I added trimming of escape sequences for compatibility with Windows.

Preview:
image

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -180,6 +180,10 @@ prompt_pure_precmd() {
 	# Check if we should display the virtual env, we use a sufficiently high
 	# index of psvar (12) here to avoid collisions with user defined entries.
 	psvar[12]=
+	# Check if a conda environment is active and display it's name
+	if [[ -n $CONDA_DEFAULT_ENV ]]; then
+		psvar[12]="${CONDA_DEFAULT_ENV//[$'\t\r\n']}"
+	fi
 	# When VIRTUAL_ENV_DISABLE_PROMPT is empty, it was unset by the user and
 	# Pure should take back control.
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -186,6 +186,11 @@ prompt_pure_precmd() {
 		psvar[12]="${VIRTUAL_ENV:t}"
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
+	# Check if a conda environment is active and display it's name
+	if [[ -n $CONDA_DEFAULT_ENV ]]; then
+		psvar[12]="${CONDA_DEFAULT_ENV//[$'\t\r\n']}"
+		export VIRTUAL_ENV_DISABLE_PROMPT=12
```
mafredri: Why set `VIRTUAL_ENV_DISABLE_PROMPT` here? It has no effect on conda. Also, what if both conda and virtualenv are active at the same time? In that case, the conda `psvar[12]` would override the virtualenv which could be confusing.

sbmueller:
Hi @mafredri 

>We have no way to tell conda not to modify the prompt (PS1), conda will do this change, and it might mess with Pure in one way or another (this is not the situation with virtualenv)

The user can disable this with `conda config --set changeps1 False`. However an automated solution would be nicer. But IMHO this is not too bad.

>We need to consider the implications of both virtualenv and conda being active at the same time, does it even work?

I am no virtualenv user, but I would suggest both are just prepending `$PATH`? This would imply that whatever environment was activated the latest will overwrite the one before. I will try to have a look here.

pure.zsh
sbmueller: You're right, `VIRTUAL_ENV_DISABLE_PROMPT` should not be altered.
```suggestion
```

mafredri:
Thanks again 🎉!

PR #450, opened 2018-12-03
Fix username color

LitoMore:
The white color is not compatible with the Terminal which has a white background. We could use other colors. I think the `reset` is a good choice.


LitoMore:
@mafredri No problem, I suggested this change at here https://github.com/sindresorhus/pure/pull/472#pullrequestreview-248311292.

PR #454, opened 2018-12-14
Update to zsh-async 1.7.1 and recover from unexpected worker death

symphorien:
mafredri/zsh-async#31 fixes a bug which has affected me where in some conditions the worker dies unexpectedly. zsh-async now recovers gracefully from this condition by tearing the worker down and notifying the callback.
This commit updates zsh-async and add some code to handle worker disappearance. The worker will be restarted (if I understand correctly) on next prompt redraw. I tested by waiting for the worker to die and then `cd`ing in another directory and checking that the git information updates correctly.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -386,6 +386,10 @@ prompt_pure_async_callback() {
 	local do_render=0
 
 	case $job in
+		\[async])
+			# our worker died unexpectedly
+			typeset -g prompt_pure_async_init=0
+			;;
 		prompt_pure_async_vcs_info)
 			local -A info
 			typeset -gA prompt_pure_vcs_info
--- a/pure.zsh
+++ b/pure.zsh
@@ -391,8 +391,11 @@ prompt_pure_async_callback() {
 
 	case $job in
 		\[async])
-			# our worker died unexpectedly
-			typeset -g prompt_pure_async_init=0
+			# code is 1 for corrupted worker output and 2 for dead worker
+			if [[ $code -eq 2 ]]; then
+				# our worker died unexpectedly
+				typeset -g prompt_pure_async_init=0
+			fi
 			;;
 		prompt_pure_async_vcs_info)
 			local -A info
--- a/async.zsh
+++ b/async.zsh
@@ -3,12 +3,12 @@
 #
 # zsh-async
 #
-# version: 1.7.0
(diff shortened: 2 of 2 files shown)


pure.zsh
```diff
@@ -390,6 +390,10 @@ prompt_pure_async_callback() {
 	local do_render=0
 
 	case $job in
+		\[async])
```
mafredri: I think we need a way to differentiate between the two types of `[async]` errors that can be received here. One is https://github.com/mafredri/zsh-async/blob/98d32afcce8e328be86f381c529e951943e57d19/async.zsh#L288 which returns an exit code of 1 (albeit that behavior is undefined). And then there's the new error (https://github.com/mafredri/zsh-async/blob/98d32afcce8e328be86f381c529e951943e57d19/async.zsh#L320) that indicates the worker has been killed. Just so happens it returns an exit code of 2 now, also undefined.

We could rename the new error to `[async/crash]`. Or we could start defining the error codes. For now checking for `2` should be sufficient though.

pure.zsh
```diff
@@ -390,6 +390,10 @@ prompt_pure_async_callback() {
...
+			# our worker died unexpectedly
+			typeset -g prompt_pure_async_init=0
```
mafredri: The worker could have crashed during some async invocation after the prompt has rendered. In this case, just setting init to zero will not recreate the worker and re-initiate the tasks. Not sure if we want to handle it or not, but I think we do. Worst case the user is left with an empty prompt until they hit enter (e.g. missing git information).

symphorien: I'd say this is not a problem. Rationale: pure is async as in, if information is currently unavailable, then bail out. Here is a new case where information is unavailable, so bail out. On the next prompt draw, we will retry. This seems sane to me.

Also, this keeps complexity low. I am afraid doing better would be fairly complex for a scenario which is very infrequent at best.

symphorien:
Anything I should add ?

mafredri:
@symphorien not that I can think of. I've been running this for a while now and I haven't run into any issues. Not that I have triggered the issue that is fixed by this PR either, but still 😅.

Thanks!

sindresorhus:
@mafredri Should we do another release? It's been a while since the last release and there are some useful commits on master.

PR #458, opened 2019-01-27
Add pure-now to Ports section in the readme

chabou:
`pure-now` is a maintained fork with Zeit's `now` support

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -187,6 +187,7 @@ See [FAQ Archive](https://github.com/sindresorhus/pure/wiki/FAQ-Archive) for pre
 	- [therealklanni/purity](https://github.com/therealklanni/purity) - More compact current working directory, important details on the main prompt line, and extra Git indicators.
  	- [intelfx/pure](https://github.com/intelfx/pure) - Solarized-friendly colors, highly verbose, and fully async Git integration.
 	- [dfurnes/purer](https://github.com/dfurnes/purer) - Compact single-line prompt with built-in Vim-mode indicator.
+	- [chabou/pure-now](https://github.com/chabou/pure-now) - Maintained fork with [Now](https://zeit.co/now) support
 - **Bash**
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -187,6 +187,7 @@ See [FAQ Archive](https://github.com/sindresorhus/pure/wiki/FAQ-Archive) for pre
 	- [therealklanni/purity](https://github.com/therealklanni/purity) - More compact current working directory, important details on the main prompt line, and extra Git indicators.
  	- [intelfx/pure](https://github.com/intelfx/pure) - Solarized-friendly colors, highly verbose, and fully async Git integration.
 	- [dfurnes/purer](https://github.com/dfurnes/purer) - Compact single-line prompt with built-in Vim-mode indicator.
+	- [chabou/pure-now](https://github.com/chabou/pure-now) - Maintained fork with [Now](https://zeit.co/now) support
```
mafredri: Could we remove the `Maintained`-part? Happy to merge this but I don't want us to make any claims if a fork is maintained or not 😄.

chabou: Yes you're right :)

mafredri:
Thanks!

sindresorhus:
@mafredri Could we add hooks somehow so people could add stuff like this without having to maintain a fork?

mafredri:
@sindresorhus, yeah definitely. It’s something I’ve been thinking about for a long time. Just haven’t gotten around to commit to it / figure out the best aporoach to do it.

In a way it’s multi factored. It would be nice to provide an API that allows user defined segments to run asynchronously as well. And it should be simple enough that it’s hard to make mistakes. Ordering of parts/segments would also be nice and if we ever make a new segment it would be nice if the user doesn’t have to manually enable it due to hard coding the current segments, or whatnot.

Perhaps functions like `prompt_pure_add_segment myfunc` and `_set_order myfunc 3` (moves it to pos 3 and old 3-n are offset by one).

Another thought I’ve had is more hook-like, e.g. right before re-rendering the prompt. A hook function is called where you can manually modify the preprompt array, but it requires a bit more intimate knowledge of how to work with zsh arrays.

sindresorhus:
@mafredri Could you open a new issue about this (copying your above comment) do we can continue the discussion there?

PR #459, opened 2019-02-01
Skip grep fork, use native zsh matching

xPMo:
Method taken from http://zdharma.org/Zsh-100-Commits-Club/Zsh-Native-Scripting-Handbook.html

Just a small optimization, I obsess over this kind of stuff. Thanks for pure btw, I've adapted it quite a bit for my [own purposes.](/xPMo/twodir-zsh-theme)

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -517,7 +517,9 @@ prompt_pure_state_setup() {
 		who_out=$(who -m 2>/dev/null)
 		if (( $? )); then
 			# Who am I not supported, fallback to plain who.
-			who_out=$(who 2>/dev/null | grep ${TTY#/dev/})
+			local -a who_in
+			who_in=( ${(f)"$(who 2>/dev/null)"} )
+			who_out="${(M)who_in:#*[[:space:]]${TTY#/dev/}[[:space:]]*}"
 		fi
 
 		local reIPv6='(([0-9a-fA-F]+:)|:){2,}[0-9a-fA-F]+'  # Simplified, only checks partial pattern.


pure.zsh
```diff
@@ -517,7 +517,9 @@ prompt_pure_state_setup() {
 		who_out=$(who -m 2>/dev/null)
 		if (( $? )); then
 			# Who am I not supported, fallback to plain who.
-			who_out=$(who 2>/dev/null | grep ${TTY#/dev/})
+			declare -a who_in
```
mafredri: This is a bit pedantic but could we change `declare` to `local` here? That, or `typeset`. The reason I'm suggesting it is that Pure doesn't use `declare` anywhere else, and for anyone reading this code, the mixture of declarations might just cause confusion. I.e. why use `declare` here, does it have a special purpose? I personally like `local` due to the additional information the name conveys (this is a local variable) albeit using `typeset` (or `declare` for that matter) in this context produces the exact same result.

Apologies for the wall of text 😅.

pure.zsh
```diff
@@ -517,7 +517,9 @@ prompt_pure_state_setup() {
...
+			who_in=( ${(f)"$(who 2>/dev/null)"} )
+			who_out="${(M)who_in:#*${TTY#/dev/}*}"
```
mafredri: I noticed a bug in the previous implementation which we could solve by changing the matching ever so slightly => `*${TTY#/dev/}[^0-9]*`. I did a quick test and this seems to work, can you find any problems with it or a better approach?

So the problem is that `pts/1` matches `pts/12` (as an example), in both this PR and previous implementation, this would produce multiple matches which is not what we want.

pure.zsh
xPMo: Whoops, I thought there was something going on with the `local` keyword and `-a`, but apparently I misread `-g` as `-a` in the manpage.  Committing.

pure.zsh
xPMo: `*[:space:]${TTY#/dev/}[:space:]*` would be safe to the point of paranoid, which is always good.

mafredri:
Looks good, go ahead 👍.

pure.zsh
```diff
@@ -517,7 +517,9 @@ prompt_pure_state_setup() {
 		who_out=$(who -m 2>/dev/null)
 		if (( $? )); then
 			# Who am I not supported, fallback to plain who.
-			who_out=$(who 2>/dev/null | grep ${TTY#/dev/})
+			local -a who_in
+			who_in=( ${(f)"$(who 2>/dev/null)"} )
+			who_out="${(M)who_in:#*[:space:]${TTY#/dev/}[:space:]*}"
```
mafredri: Hmm, `[:space:]` doesn't seem to work in my shell, does it require some option to be enabled?

mafredri: Ah, seems to work with `[[:space:]]`.

xPMo: Ah, found it in the manual:

> Note that the square brackets are additional to those enclosing the whole set of characters,  so  to  test  for  a single alphanumeric character you need `[[:alnum:]]`.  Named character sets can be used alongside other types, e.g. `[[:alpha:]0-9]`.

mafredri:
@xPMo went ahead and merged this although I don't have a busybox system to test this with at the moment, but I can't think of anything that could go wrong.

Thanks again! 😄

Issue #365, opened 2017-11-13
git stash


jpresley23:
I read the discussion in Issue #317 and understand the concern to not have a messy prompt. The right prompt suggestion is a decent workaround but the problem is that shows a message when there are no changes in stash. For a light user of stash, the constant 0 in the right hand side of the screen is a distraction. Changing that requires building a function that indicates it may be more appropriate.

I use stash to store work in progress changes that I don't want to commit before I re-integrate master into my branch. It's useful to see an indicator if there are changes in stash I forgot about if I get interrupted in my workflow. I used a custom prompt that someone at work shared around that used something like `*` to indicate there is something in your stash.

Would it be a good compromise to add the stash indicator functionality but set a default minimum threshold before there is an indicator that could be changed with an environment variable? For example, the default could be a high number, say. Those who are light stash users could override that and set it to 0.

The alternative is to create a super light "pure-stash" indicator that is separate from pure.

jpresley23:
@leoj3n I'm curious on your thoughts since you opened the original issue.

henrywallace:
Alternatively, is it possible to allow calling some custom function for displaying git information? Where I still want to use the rest of pure prompt, but insert my own git information.

KlausHartl4242:
I'd love to see support for stashes... at the moment I've hacked something into `prompt_pure_preprompt_render()` on my own machine, the result looking like:

<img width="451" alt="screenshot 2019-02-08 at 19 05 37" src="https://user-images.githubusercontent.com/40851765/52496940-12814180-2bd5-11e9-856d-f5d1381ca04e.png">

If there are any stashes, the yellow ≡ icon is shown, otherwise not... (similar to how in and outgoing arrows behave) 

(I would be happy to prepare a PR for this.)

PS: Icon and color is inspired by https://github.com/michaeldfallen/git-radar, which I recently migrated from to pure...

carhartl:
I'd love to see support for stashes... at the moment I've hacked something into `prompt_pure_preprompt_render()` on my own machine, the result looking like:

<img width="451" alt="screenshot 2019-02-08 at 19 05 37" src="https://user-images.githubusercontent.com/21918/52497299-fd58e280-2bd5-11e9-8404-ff6e97dfcbb1.png">

If there are any stashes, the yellow ≡ icon is shown, otherwise not... (similar to how in and outgoing arrows behave)

(I would be happy to prepare a PR for this.)

[...]

Issue #317, opened 2017-04-20
Git stash support


leoj3n:
Would be cool if the prompt showed an indicator when the current repo has stashes.

mafredri:
How would you display it?

I'm not a big fan of the suggestion. I usually have something in my git stash and would be annoyed by the constant reminder as I don't use it for storing important stuff, it's a temporary workspace.

Just for pure I have 38 stashes, first one from `Mon Feb 23 15:33:22 2015` 😝.

leoj3n:
Just an indicator like $ or something. Maybe it would remind you to take
care of clearing

leoj3n:
![](https://media.giphy.com/media/5xaOcLCBzBw4QrtdDP2/giphy.gif)

mafredri:
> Maybe it would remind you to take care of clearing

For me it would be more of a nag to get rid of my stash, which I don't want. Most of it is junk, but I keep some things in there for a reason. E.g. dead-ends I might to re-visit or otherwise keep handy in case I need it in the future.

sindresorhus:
I'm also 👎

leoj3n:
> dead-ends I might to re-visit

Wouldn't that make more sense as a local branch?

mafredri:
> Wouldn't that make more sense as a local branch?

Some things aren't worth the effort to turn into a branch. Either way, stash is part of my workflow and I have no plans on changing how I use it 🙂.

I'm sorry if this inconveniences you @leoj3n, but I don't think we're going to implement this. This issue has now been open for 2.5 months without any up-votes. There is also no benefit to keeping a clean stash, so I see reminding people as a negative. I've already voiced my other concerns so I wont repeat those.

leoj3n:
I guess it would annoy some people, but it would remind me not to blow away important changes! 😅

> I'm sorry if this inconveniences you

No worries, not having a git stash prompt indicator won't inconvenience me that much (in turn, I'm sorry if this feature suggestion inconvenienced you!)

> There is also no benefit to keeping a clean stash

I didn't make this issue to argue that there were any benefits to "keeping a clean stash"; I just thought that was something you were struggling with, based on what you posted @mafredri ...

[...]

Issue #460, opened 2019-02-04
Add support for modifying the pre-prompt (add / edit / remove pre-prompt parts)


mafredri:
This issue is for tracking discussion and development of an extension to the pre-prompt that will allow users to make modifications to it by some means.

The purpose is to minimize the need for our users to maintain a fork just so that they can add some information to the pre-prompt.

**Prologue:** It would be nice to provide an API that allows user defined segments to run asynchronously as well. And it should be simple enough that it’s hard to make mistakes. Ordering of parts/segments would also be nice and if we ever make a new segment it would be nice if the user doesn’t have to manually enable it due to hard coding the current segments, or whatnot.

Perhaps functions like `prompt_pure_add_segment myfunc` and `_set_order myfunc 3` (moves it to pos 3 and old 3-n are offset by one).

Another thought I’ve had is more hook-like, e.g. right before re-rendering the prompt. A hook function is called where you can manually modify the preprompt array, but it requires a bit more intimate knowledge of how to work with zsh arrays.

_Issue originally raised by @sindresorhus in https://github.com/sindresorhus/pure/pull/458#issuecomment-459633908_

CircleCode:
do you expect users to list segments they would love to see available (so that you can pre-code them, even if they are not enabled by default)?
On my side, I'd love to see
* numbre of stash
* upstream name

In case you are not expecting input like this, feel free to delete my comment :-)

warbaque:
for workaround I added
`preprompt_parts+=($(custom_prompt_parts 2>/dev/null))` to `prompt_pure_preprompt_render()` with no other customization options (I just needed way to add extra parts to the end)

Then just have `custom_prompt_parts()` in my settings env

onyxraven:
Just throwing in my support - I've been using the https://github.com/intelfx/pure branch (which is now really far behind), but allows customization by adding to `prompt_pure_pieces` (https://github.com/intelfx/pure#example-1)

(I've added things such as a shrunken path via oh-my-zsh `shrink_path`, kubernetes context via `kctx`, aws-vault profile, etc.)

awmottaz:
I would also like to see this feature added. I enjoy using [`t`](https://github.com/unu/t) to manage quick todo tasks, and I would like to add the number of incomplete tasks to my prompt.

IssuehuntBot:
@issuehunt has funded $80.00 to this issue.

---
[...]

PR #462, opened 2019-02-11
Add Git stash support

carhartl:
This is adding support for git stashes. E.g. if there are any stashes show an icon, otherwise not (similar to how git arrows behave). Said icon is configurable via `PURE_GIT_STASH_SYMBOL` environment variable.

It's implemented utilizing `vcs_info`'s `misc` message + hook for tracking information about stashes.

Please take this as a proposal I guess; while #365 is still open, I'm aware of the discussion in #317..

screenshot 2019-02-11 at 09 41 54

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -155,7 +155,7 @@ prompt_pure_preprompt_render() {
 	fi
 	# Git stash symbol.
 	if [[ -n $prompt_pure_vcs_info[stash] ]]; then
-		preprompt_parts+=('%F{cyan}≡%f')
+		preprompt_parts+=('%F{cyan}${PURE_GIT_STASH_SYMBOL:-≡}%f')
 	fi
 
 	# Username and machine, if applicable.
--- a/pure.zsh
+++ b/pure.zsh
@@ -394,14 +394,14 @@ prompt_pure_async_refresh() {
 
 	if [[ -z $prompt_pure_git_fetch_pattern ]]; then
 		# We set the pattern here to avoid redoing the pattern check until the
-		# working three has changed. Pull and fetch are always valid patterns.
+		# working tree has changed. Pull and fetch are always valid patterns.
 		typeset -g prompt_pure_git_fetch_pattern="pull|fetch"
 		async_job "prompt_pure" prompt_pure_async_git_aliases
 	fi
 
 	async_job "prompt_pure" prompt_pure_async_git_arrows
 
-	# Do not preform `git fetch` if it is disabled or in home folder.
+	# Do not perform `git fetch` if it is disabled or in home folder.
 	if (( ${PURE_GIT_PULL:-1} )) && [[ $prompt_pure_vcs_info[top] != $HOME ]]; then
 		# Tell the async worker to do a `git fetch`.
 		async_job "prompt_pure" prompt_pure_async_git_fetch
--- a/pure.zsh
+++ b/pure.zsh
@@ -26,7 +26,7 @@
 
 +vi-git-stash() {
 	local -a stashes
-	stashes=$(git stash list 2>/dev/null | wc -l | tr -d ' ')
+	stashes=$(git stash list 2>/dev/null | wc -l)
 	if [[ $stashes -gt 0 ]]; then
 		hook_com[misc]="stash=${stashes}"
 	fi
--- a/pure.zsh
+++ b/pure.zsh
@@ -271,8 +271,7 @@ prompt_pure_async_vcs_info() {
 	# and stash information via misc (%m).
 	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
 	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a' '%m'
-	zstyle -t ":prompt:pure:git:stash" show
-	if [[ $? == 0 ]]; then
+	if [[ $1 == 0 ]]; then
 		zstyle ':vcs_info:git*+set-message:*' hooks git-stash
 	fi
 
@@ -381,7 +380,8 @@ prompt_pure_async_tasks() {
 	fi
 	unset MATCH MBEGIN MEND
 
-	async_job "prompt_pure" prompt_pure_async_vcs_info
+	zstyle -t ":prompt:pure:git:stash" show
+	async_job "prompt_pure" prompt_pure_async_vcs_info $?
 
 	# Only perform tasks inside a Git working tree.
 	[[ -n $prompt_pure_vcs_info[top] ]] || return
--- a/pure.zsh
+++ b/pure.zsh
@@ -155,7 +155,7 @@ prompt_pure_preprompt_render() {
 	fi
 	# Git stash symbol.
 	if [[ -n $prompt_pure_vcs_info[stash] ]]; then
-		preprompt_parts+=('%F{cyan}${PURE_GIT_STASH_SYMBOL:-≡}%f')
+		preprompt_parts+=('%F{$prompt_pure_colors[git:stash]}${PURE_GIT_STASH_SYMBOL:-≡}%f')
 	fi
 
 	# Username and machine, if applicable.
@@ -699,6 +699,7 @@ prompt_pure_setup() {
 	prompt_pure_colors_default=(
 		execution_time       yellow
 		git:arrow            cyan
+		git:stash            cyan
 		git:branch           242
 		git:branch:cached    red
 		git:action           242
--- a/pure.zsh
+++ b/pure.zsh
@@ -24,6 +24,14 @@
 # \e[2K => clear everything on the current line
 
 
++vi-git-stash() {
+	local -a stashes
+	stashes=$(git stash list 2>/dev/null | wc -l | tr -d ' ')
+	if [[ $stashes -gt 0 ]]; then
+		hook_com[misc]="stash=${stashes}"
+	fi
+}
+
 # Turns seconds into human readable time.
 # 165392 => 1d 21h 56m 32s
 # https://github.com/sindresorhus/pretty-time-zsh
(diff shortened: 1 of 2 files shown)


sindresorhus:
If we decide to accept this, it should at minimum be opt-in [using `zstyle` as outlined in](https://github.com/sindresorhus/pure/issues/306#issuecomment-292542352) (someone needs to implement that).

Alternatively, https://github.com/sindresorhus/pure/issues/460 could allow you do add this yourself without having to fork Pure.

sindresorhus:
@mafredri Are you ok with adding this as opt-in through zstyle?

mafredri:
> @mafredri Are you ok with adding this as opt-in through zstyle?

I think that would be alright. So would we have an env variable for configuring the icon, and a zstyle for enabling the feature? Or should we take this opportunity to start defining icons in zstyle as well?

sindresorhus:
> Or should we take this opportunity to start defining icons in zstyle as well?

Yes

sindresorhus:
@carhartl Can you make it use zstyle?

carhartl:
@sindresorhus Need to learn more about that but happy to try...

carhartl:
@sindresorhus  So it‘s supposed to be implemented as @mafredri suggested:

zstyle for enabling the feature
zstyle for defining the icon

?

carhartl:
I think it would be better to migrate all of the icons to zstyle, not just the one for stashes. This seems out of scope of this PR and I'd like to do this in a follow-up PR.

sindresorhus:
> So it‘s supposed to be implemented as @mafredri suggested:
zstyle for enabling the feature
zstyle for defining the icon

👍 

> I think it would be better to migrate all of the icons to zstyle, not just the one for stashes. This seems out of scope of this PR and I'd like to do this in a follow-up PR.

👍

carhartl:
@sindresorhus

Showing the stash status does now require opt-in via `zstyle ":prompt:pure:git:stash:show" enable yes`. The stash symbol color is also zstyle'd.

Is there anything else needed for this to be merged?

carhartl:
Hmmm, might be nicer if we‘re not setting up the hook in the first place unless opted in!

readme.md
```diff
@@ -70,7 +70,11 @@ prompt pure
 | **`PURE_PROMPT_VICMD_SYMBOL`**   | Defines the prompt symbol used when the `vicmd` keymap is active (VI-mode).                    | `❮`            |
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
+| **`PURE_GIT_STASH_SYMBOL`**      | Defines the git stash symbol.                                                                  | `≡`            |
 
+Showing git stash status as part of the prompt is not activated by default. To activate this you'll need to opt in via `zstyle`:
+
+`zstyle ':prompt:pure:git:stash:show' enable yes`
```
mafredri: Show and enable sounds a bit repetitive to me. Could it make sense to drop `:show` here? Or drop show and rename enable => show.

pure.zsh
```diff
@@ -253,19 +266,22 @@ prompt_pure_async_vcs_info() {
...
-	# Export branch (%b), Git toplevel (%R), and action (rebase/cherry-pick) (%a).
-	zstyle ':vcs_info:git*' formats '%b' '%R'
-	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a'
+	# Only export four message variables from `vcs_info`.
+	zstyle ':vcs_info:*' max-exports 4
+	# Export branch (%b), Git toplevel (%R), action (rebase/cherry-pick) (%a),
+	# and stash information via misc (%m).
+	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
+	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a' '%m'
+	zstyle ':vcs_info:git*+set-message:*' hooks git-stash
```
mafredri: Ideally I’d like us to avoid “doing the work” for gut stash, unless it’s enabled. But it’s not critical for this PR and we cam change it afterwards.

pure.zsh
```diff
@@ -253,19 +266,22 @@ prompt_pure_async_vcs_info() {
...
-	# Only export three message variables from `vcs_info`.
-	zstyle ':vcs_info:*' max-exports 3
-	# Export branch (%b), Git toplevel (%R), and action (rebase/cherry-pick) (%a).
-	zstyle ':vcs_info:git*' formats '%b' '%R'
-	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a'
+	# Only export four message variables from `vcs_info`.
+	zstyle ':vcs_info:*' max-exports 4
+	# Export branch (%b), Git toplevel (%R), action (rebase/cherry-pick) (%a),
+	# and stash information via misc (%m).
+	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
```
mafredri: I’m guessing there are no repercussions from including `%a` here?

sindresorhus:
I think we should is the wording `enabled` instead of `show`, so we can use the same wording for features that are not visible ones.

sindresorhus:
Can we make it `zstyle ':prompt:pure:git:stash' enable yes`? (Without the `show`)

mafredri:
Looks like we all had a few of the same ideas 😅

carhartl:
> Can we make it zstyle ':prompt:pure:git:stash' enable yes?

`:prompt:pure:git:stash` is used for the color already, to prevent that clash I thought I need to introduce the additional `:show`. So with @mafredri's suggestion, would 

`zstyle ':prompt:pure:git:stash:show' show yes`

be ok? Or maybe `zstyle ':prompt:pure:git' stash yes` is more in line with the intended usage of zstyle?

mafredri:
> `:prompt:pure:git:stash` is used for the color already, to prevent that clash I thought I need to introduce the additional `:show`. So with @mafredri's suggestion, would

That’s ok because there can be multiple values under a `zstyle`. So `:prompt:pure:git:stash` should be able to store both color=x and show=y. 

> be ok? Or maybe `zstyle ':prompt:pure:git' stash yes` is more in line with the intended usage of zstyle?

This could also work. For me the former feels a bit more uniform but this is essentially a matter of enabling / disabling showing of git features so `¯\_(ツ)_/¯`. I’m fine with either.

carhartl:
Ah, please bear with me... I missed to think of `:prompt:pure:git:stash` as some sort of namespace. Thanks for clarifying.

This changes everything :)

Then I also think we should go for `:prompt:pure:git:stash`, will allow to configure related things in the same way, as pointed out.

```
zstyle ':prompt:pure:git:stash' show yes
zstyle ':prompt:pure:git:stash' color yellow
zstyle ':prompt:pure:git:stash' icon S
```

Updating the commit...

pure.zsh
carhartl: For some reason it was producing an error when `formats` didn't have the same amount as `actionformats`. I really don't understand why these two seem to interact, maybe I did something wrong else, still new to zsh internals.

pure.zsh
```diff
@@ -271,7 +271,10 @@ prompt_pure_async_vcs_info() {
 	# and stash information via misc (%m).
 	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
 	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a' '%m'
-	zstyle ':vcs_info:git*+set-message:*' hooks git-stash
+	zstyle -t ":prompt:pure:git:stash" show
```
mafredri: One thing to keep in mind is that since this is running inside the async worker. Retroactive changes to zstyle won’t propagate to here. A better option might be to pass a boolean argument (1/0) to this function when it’s invoked (as async job)

pure.zsh
mafredri: Alright, it’s been a while since I worked with vcs_info as well so just being cautious 😄. It’s probably fine.

carhartl: Took a second look... `%m` was only required in `formats` only, but not in `actionformats`.

mafredri: Interesting. How about when a git rebase is active? That’s when action formats kick in AFAICR.

carhartl: And the actionformats are for when git is performing a merge or rebase. If I understand all this correctly, if we were omitting `%m` from actionformats, `info[stash]=$vcs_info_msg_3_` would remain empty (?) and the stash symbol would never be shown during such an action. Which is what we want?

carhartl: I guess we need to ensure that `$vcs_info_msg_3_` would always point to the expected output...

pure.zsh
carhartl: Would that be in that line:

```
async_job "prompt_pure" prompt_pure_async_vcs_info
```

?

mafredri: That’s right. Read the value before and then pass it along there :).

carhartl: Fixed..

carhartl:
Is there anything you need me to add?

fladson:
I would love to see this merged, thanks for adding it.

readme.md
```diff
@@ -70,7 +70,11 @@ prompt pure
 | **`PURE_PROMPT_VICMD_SYMBOL`**   | Defines the prompt symbol used when the `vicmd` keymap is active (VI-mode).                    | `❮`            |
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
+| **`PURE_GIT_STASH_SYMBOL`**      | Defines the git stash symbol.                                                                  | `≡`            |
```
sindresorhus: As discussed in the comments, this should be `zstyle ':prompt:pure:git:stash' icon S`.

sindresorhus:
@carhartl Sorry, this fell off my radar. There are some remaining feedback: 

- https://github.com/sindresorhus/pure/pull/462#discussion_r352346278
- https://github.com/sindresorhus/pure/pull/462#discussion_r376733428

carhartl:
@sindresorhus 

Oh, ok. Regarding the icon configuration I thought you agreed that this would be out of scope and would better be done for all icons instead of just this new one in a separate PR.

See https://github.com/sindresorhus/pure/pull/462#issuecomment-560091472

sindresorhus:
Forgot about my comment. Never mind the zconfig thing then.

carhartl:
Regarding the other issue: I changed it to

`zstyle ':prompt:pure:git:stash' show yes`

but it looks like I overlooked to adapt that in one place in the readme.

carhartl:
@sindresorhus I have fixed the issue in the readme...

sindresorhus:
Looks good. Thanks for working on this  🙌

PR #464, opened 2019-02-26
Allow setting pre-prompt by user ENV var

samcv:
This will allow you to add stuff before the start of the prompt and/or
change the color of the path.

For example to change the color of the path the user can do:
PURE_PROMPT_PRE='%F{blue}'

To add a clock before the line and alter the color of the path:
PURE_PROMPT_PRE='%* %F{blue}'

Other things can be done but this gives the user a lot of control to
customize the prompt.

Let me know your thoughts, and if it makes more sense to split the
path color and the PRE into separate variables.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -113,9 +113,9 @@ prompt_pure_preprompt_render() {
 
 	# Initialize the preprompt array.
 	local -a preprompt_parts
-
+	local default_prompt_pre='%F{blue}'
 	# Set the path.
-	preprompt_parts+=('%F{blue}%~%f')
+	preprompt_parts+=(${PURE_PROMPT_PRE:-$default_prompt_pre}'%~%f')
 
 	# Add git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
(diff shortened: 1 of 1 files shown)


sindresorhus:
The coloring is already being fixed in https://github.com/sindresorhus/pure/pull/472.

Are there any other use-cases for this?

samcv:
@sindresorhus what I use this for is to add a clock at the left side of the prompt, as well as change the color of the directory.

So it would be nice if we could add additional things at the left hand side.

ashb:
Something like the `pure_pieces` from https://github.com/intelfx/pure#custom-handlers would support this:

```zsh
# .zshrc

# optionally define some options
PURE_CMD_MAX_EXEC_TIME=10

# optionally define custom generators
prompt_custom_time() {
	preprompt+=( "%F{gray}[%T]%f" )
}

prompt pure

# add the generator where it's needed
prompt_pure_pieces=(
	prompt_custom_time
	${prompt_pure_pieces:1}
)
```

@sindresorhus Would adding something like this be welcomed in a PR here?

ashb:
See also #460

sindresorhus:
Closing as this needs to be more fully thought through in #460. We also don't use environment variables anymore.

sindresorhus:
@ashb Can you comment in that issue instead?

PR #472, opened 2019-05-05
Add support for configuring colors with zstyle

### LINKED ISSUE #126: Excessively gray pre-prompt with ssh and git
The pre-prompt becomes excessively gray (too little contrast) when inside a git directory over an ssh connection:

Looks OK as root though:

### LINKED ISSUE #303: Feature request | arrow prompt ❯ color
The arrow prompt `❯` color for normal (`$? -eq 0`) and not normal (`$? -ne 0`) is not very obvious. I suggest you use grey or green for the arrow prompt when normal (`$? -eq 0`).

### LINKED ISSUE #306: Proposal: configure Pure colors via prompt init

### LINKED ISSUE #389: How to set color for path and git info?
The default colors are really dark for by terminal background. After searching for a while in the issues and making some tries I wasn't able to figure it out.

Thanks in advance and thank you for the awesome project.

bricewge:
Following the discussion at #306 this PR add the capacity to change the colors of `pure` by using `zstyle` to avoid polluting the global name space.

Available colors:
- branch
- error
- exec_time
- git_arrow
- host
- path
- root
- success
- user
- virtualenv

Fixes #126
Fixes #303 
Fixes #306 
Fixes #389 
Closes #256 
Closes #450

EDIT: Add requested colors.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -106,8 +106,8 @@ prompt_pure_preexec() {
 # Change the colors if their value are different from the current ones
 prompt_pure_set_colors() {
 	local color_temp
-	for key val in ${(kv)prompt_pure_colors}; do
-		zstyle -t ":prompt:pure:$key" color "$val"
+	for key value in ${(kv)prompt_pure_colors}; do
+		zstyle -t ":prompt:pure:$key" color "$value"
 		case $? in
 			1) # The current style is different from the one from zstyle
 				zstyle -s ":prompt:pure:$key" color color_temp
--- a/pure.zsh
+++ b/pure.zsh
@@ -103,19 +103,34 @@ prompt_pure_preexec() {
 	export VIRTUAL_ENV_DISABLE_PROMPT=${VIRTUAL_ENV_DISABLE_PROMPT:-12}
 }
 
+# Change the colors if their value are different from the current ones
+prompt_pure_set_colors() {
+	local color_temp
+	for key val in ${(kv)prompt_pure_colors}; do
+		zstyle -t ":prompt:pure:$key" color "$val"
+		case $? in
+			1) # current style is diffrent than the one from zstyle
+				zstyle -s ":prompt:pure:$key" color color_temp
+				prompt_pure_colors[$key]=$color_temp ;;
+			2) # no style is defined
+				prompt_pure_colors[$key]=$prompt_pure_colors_default[$key] ;;
+		esac
+	done
+}
+
 prompt_pure_preprompt_render() {
 	setopt localoptions noshwordsplit
 
 	# Set color for git branch/dirty status, change color if dirty checking has
 	# been delayed.
-	local git_color=242
-	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=red
+	local git_color=$prompt_pure_colors[git:branch]
+	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=$prompt_pure_colors[git:branch:cached]
 
 	# Initialize the preprompt array.
 	local -a preprompt_parts
 
 	# Set the path.
-	preprompt_parts+=('%F{blue}%~%f')
+	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%~%f')
 
 	# Add git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
@@ -124,13 +139,13 @@ prompt_pure_preprompt_render() {
 	fi
 	# Git pull/push arrows.
 	if [[ -n $prompt_pure_git_arrows ]]; then
-		preprompt_parts+=('%F{cyan}${prompt_pure_git_arrows}%f')
+		preprompt_parts+=('%F{$prompt_pure_colors[git:arrow]}${prompt_pure_git_arrows}%f')
 	fi
 
 	# Username and machine, if applicable.
 	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=('${prompt_pure_state[username]}')
 	# Execution time.
-	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{yellow}${prompt_pure_cmd_exec_time}%f')
+	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{$prompt_pure_colors[exec_time]}${prompt_pure_cmd_exec_time}%f')
 
 	local cleaned_ps1=$PROMPT
 	local -H MATCH MBEGIN MEND
@@ -173,6 +188,9 @@ prompt_pure_precmd() {
 
 	# shows the full path in the title
 	prompt_pure_set_title 'expand-prompt' '%~'
+
+	# Modify the colors if some have changed
+	prompt_pure_set_colors
 
 	# preform async git dirty check and fetch
 	prompt_pure_async_tasks
@@ -509,7 +527,7 @@ prompt_pure_state_setup() {
 
 	# Check SSH_CONNECTION and the current state.
 	local ssh_connection=${SSH_CONNECTION:-$PROMPT_PURE_SSH_CONNECTION}
-	local username
+	local username hostname
 	if [[ -z $ssh_connection ]] && (( $+commands[who] )); then
 		# When changing user on a remote system, the $SSH_CONNECTION
 		# environment variable can be lost, attempt detection via who.
@@ -542,11 +560,12 @@ prompt_pure_state_setup() {
 		unset MATCH MBEGIN MEND
 	fi
 
+	hostname='%F{$prompt_pure_colors[host]}@%m%f'
 	# show username@host if logged in through SSH
-	[[ -n $ssh_connection ]] && username='%F{242}%n@%m%f'
+	[[ -n $ssh_connection ]] && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
 
 	# show username@host if root, with username in default color
-	[[ $UID -eq 0 ]] && username='%f%n%f%F{242}@%m%f'
+	[[ $UID -eq 0 ]] && username='%F{$prompt_pure_colors[user:root]}%n%f'"$hostname"
 
 	typeset -gA prompt_pure_state
 	prompt_pure_state[version]="1.9.0"
@@ -618,6 +637,7 @@ prompt_pure_setup() {
 	zmodload zsh/datetime
 	zmodload zsh/zle
 	zmodload zsh/parameter
+	zmodload zsh/zutil
 
 	autoload -Uz add-zsh-hook
 	autoload -Uz vcs_info
@@ -626,6 +646,23 @@ prompt_pure_setup() {
 	# The add-zle-hook-widget function is not guaranteed
 	# to be available, it was added in Zsh 5.3.
 	autoload -Uz +X add-zle-hook-widget 2>/dev/null
+
+	# Set the colors
+	typeset -gA prompt_pure_colors_default prompt_pure_colors
+	prompt_pure_colors_default=(
+		exec_time            yellow
+		git:arrow            cyan
+		git:branch           242
+		git:branch:cached    red
+		host                 242
+		path                 blue
+		prompt:error         red
+		prompt:success       magenta
+		user                 242
+		user:root            default
+		virtualenv           242
+	)
+	prompt_pure_colors=("${(@fkv)prompt_pure_colors_default}")
 
 	add-zsh-hook precmd prompt_pure_precmd
 	add-zsh-hook preexec prompt_pure_preexec
@@ -640,10 +677,10 @@ prompt_pure_setup() {
 	fi
 
 	# if a virtualenv is activated, display it in grey
-	PROMPT='%(12V.%F{242}%12v%f .)'
+	PROMPT='%(12V.%F{$prompt_pure_colors[virtualenv]}%12v%f .)'
 
 	# prompt turns red if the previous command didn't exit with 0
-	PROMPT+='%(?.%F{magenta}.%F{red})${prompt_pure_state[prompt]}%f '
+	PROMPT+='%(?.%F{$prompt_pure_colors[prompt:success]}.%F{$prompt_pure_colors[prompt:error]})${prompt_pure_state[prompt]}%f '
 
 	# Store prompt expansion symbols for in-place expansion via (%). For
 	# some reason it does not work without storing them in a variable first.
--- a/pure.zsh
+++ b/pure.zsh
@@ -109,10 +109,10 @@ prompt_pure_set_colors() {
 	for key val in ${(kv)prompt_pure_colors}; do
 		zstyle -t ":prompt:pure:$key" color "$val"
 		case $? in
-			1) # current style is diffrent than the one from zstyle
+			1) # The current style is different from the one from zstyle
 				zstyle -s ":prompt:pure:$key" color color_temp
 				prompt_pure_colors[$key]=$color_temp ;;
-			2) # no style is defined
+			2) # No style is defined
 				prompt_pure_colors[$key]=$prompt_pure_colors_default[$key] ;;
 		esac
 	done
--- a/pure.zsh
+++ b/pure.zsh
@@ -145,7 +145,7 @@ prompt_pure_preprompt_render() {
 	# Username and machine, if applicable.
 	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=('${prompt_pure_state[username]}')
 	# Execution time.
-	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{$prompt_pure_colors[exec_time]}${prompt_pure_cmd_exec_time}%f')
+	[[ -n $prompt_pure_cmd_exec_time ]] && preprompt_parts+=('%F{$prompt_pure_colors[execution_time]}${prompt_pure_cmd_exec_time}%f')
 
 	local cleaned_ps1=$PROMPT
 	local -H MATCH MBEGIN MEND
@@ -650,7 +650,7 @@ prompt_pure_setup() {
 	# Set the colors
 	typeset -gA prompt_pure_colors_default prompt_pure_colors
 	prompt_pure_colors_default=(
-		exec_time            yellow
+		execution_time       yellow
 		git:arrow            cyan
 		git:branch           242
 		git:branch:cached    red
--- a/readme.md
+++ b/readme.md
@@ -93,23 +93,25 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+
 ## Colors
 
-Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default and what part they affect are:
-- `exec_time` (yellow), the execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
-- `git:arrow` (cyan), for `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
-- `git:branch` (242), the name of the current branch when in a git repository
-- `git:branch:cached` (red), the name of the current branch when the data isn't fresh
-- `host` (242), the hostname when on a remote machine
-- `path` (blue), the current path ie. `PWD`
-- `prompt:error` (red), the `PURE_PROMPT_SYMBOL` when the previous command has *failed*
-- `prompt:success` (magenta), the `PURE_PROMPT_SYMBOL` when the previous command has *succeded*
-- `user` (242), the username when on remote machine
-- `user:root` (default), the username when the user is root
-- `virtualenv` (242), the name of the python virtualenv when in use
+Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default, and what part they affect are:
+- `exec_time` (yellow) - The execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
+- `git:arrow` (cyan) - For `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
+- `git:branch` (242) - The name of the current branch when in a Git repository
+- `git:branch:cached` (red) - The name of the current branch when the data isn't fresh
+- `host` (242) - The hostname when on a remote machine
+- `path` (blue) - The current path, for example, `PWD`
+- `prompt:error` (red) - The `PURE_PROMPT_SYMBOL` when the previous command has *failed*
+- `prompt:success` (magenta) - The `PURE_PROMPT_SYMBOL` when the previous command has *succeded*
+- `user` (242) - The username when on remote machine
+- `user:root` (default) - The username when the user is root
+- `virtualenv` (242) - The name of the Python `virtualenv` when in use
 
-The following diagram shows where each colors are applied on the prompt.
-``` text
+The following diagram shows where each color is applied on the prompt:
+
+```
 path
 |          git:branch
 |          |       git:arrow
@@ -122,6 +124,7 @@ venv ❯               |                   |
 |    prompt
 virtualenv
 ```
+
 
 ## Example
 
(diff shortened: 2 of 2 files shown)


eladchen:
Any chance this will be merged soon?

pure.zsh
```diff
@@ -543,10 +543,10 @@ prompt_pure_state_setup() {
 	fi
 
 	# show username@host if logged in through SSH
-	[[ -n $ssh_connection ]] && username='%F{242}%n@%m%f'
+	[[ -n $ssh_connection ]] && username='%F{${prompt_pure_colors[user]}}%n@%m%f'
 
 	# show username@host if root, with username in white
-	[[ $UID -eq 0 ]] && username='%F{white}%n%f%F{242}@%m%f'
+	[[ $UID -eq 0 ]] && username='%F{${prompt_pure_colors[root]}}%n%f%F{242}@%m%f'
```
cole-h: Missed a color here: `%F{242}`. Suggested: `%F{${prompt_pure_colors[host]}}` and also add above (in ssh section).

pure.zsh
```diff
@@ -598,7 +619,7 @@ prompt_pure_setup() {
 	PROMPT='%(12V.%F{242}%12v%f .)'
```
cole-h: Also missed a color here. Maybe `%F{${prompt_pure_colors[venv]}}` or `[virtualenv]`?

bricewge: It was intentional, nobody asked for it before and since the author want the project do be KISS I skipped it.

pure.zsh
bricewge: Also intentional because I didn't grasp the full intent of the authors. Blaming those lines (https://github.com/sindresorhus/pure/pull/94/files & https://github.com/sindresorhus/pure/pull/108/files) it seems the formatting was an after thought. I went with your suggestion.

pure.zsh
```diff
@@ -582,6 +583,29 @@ prompt_pure_setup() {
 	# to be available, it was added in Zsh 5.3.
 	autoload -Uz +X add-zle-hook-widget 2>/dev/null
 
+	# most used colors
+	typeset -gA prompt_pure_colors
+	prompt_pure_colors=(
```
cole-h: I would change this to 

```
	prompt_pure_colors=(
		path       blue
		branch     242
		git_arrow  cyan
		root       white
		user       242
		host       242
		exec_time  yellow
		virtualenv 242
		success    magenta
		error      red
  )
```

Just a small nit, but it makes more sense to me to have colors in the order that their sections appear (by default) on the prompt.

---

Additionally, I know it probably doesn't matter, but would you be opposed to moving the array to the top of the function? Considering there are things above it that currently reference the array, it makes more logical sense to have it be at the top despite it already being global. Would it make sense to you to do that in this situation?

bricewge: I will pass on both. The colors are sorted alphabetically because it's easier to insert them that way. The position of the array is after all the loading of all the external functions and before any actual function calls.

readme.md
```diff
@@ -93,6 +93,13 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+## Colors
+
+Colors can be changed by using `zstyle` with the pattern
+`:prompt:pure:$color_name` and style `color`. The availaible color names are:
+`branch`, `error`, `exec_time`, `git_arrow`, `host`, `path`, `root`, `success`,
+`user` and `virtualenv`.
```
sindresorhus: I think the docs here needs to be expanded. It should be clearer what the different parts actually are. I would make it a list and add a screenshot showing each component with an arrow.

Also don't hard-wrap.

bricewge: Thank you for taking the time to review this PR.

I have removed the hard-wraps and extend the documentation to explain what parts of the prompt each color affect. However I didn't made a visual explanation as you requested because GUI application are not my *forte* and I don't see how to annotate a screenshot with ten colors without overcrowding it. But if you insist I'll try.

sindresorhus: I'm happy to consider other ways to document it, but we need to make it perfectly clear what the different parts are.

bricewge: I have added a ASCII diagram that shows exactly which parts are colored by which colors. WDYT?

readme.md
```diff
@@ -93,6 +93,35 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+## Colors
+
+Colors can be changed by using `zstyle` with the pattern `:prompt:pure:$color_name` and style `color`. The availaible color names and what part they affect:
```
mafredri: Do you know if the current implementation can also support creating more specific patterns, e.g. `:prompt:pure:[state]:$color_name`? Where `[state]` could be one of `normal`, `git`, `hg`, etc.

The idea would be that we could sometime in the future add support for specifying different colors depending on e.g. what VCS system is in use. Git has a green path and Mercurial has a red one, for instance `¯\_(ツ)_/¯`.

Anyhow, I'm mainly curious. It's just an idea I've had and nothing necessary for this PR.

pure.zsh
```diff
@@ -582,6 +583,29 @@ prompt_pure_setup() {
 	# to be available, it was added in Zsh 5.3.
 	autoload -Uz +X add-zle-hook-widget 2>/dev/null
 
+	# most used colors
```
mafredri: We should add `zmodload zsh/zutil` a few lines above to make sure `zstyle` is loaded when we try to use it.

pure.zsh
```diff
@@ -582,6 +583,29 @@ prompt_pure_setup() {
...
+		virtualenv  242
+	)
+	# modify default colors if set by the user
+	local color_temp
+	for key val in ${(kv)prompt_pure_colors}; do
+		if ! zstyle -T ":prompt:pure:$key" color; then
+			zstyle -s ":prompt:pure:$key" color color_temp
+			prompt_pure_colors[$key]=$color_temp
+		fi
+	done
```
mafredri: I really like the cleanness of this implementation, however, it doesn't look like it supports changing colors at runtime and I'm concerned with how easy it will be for our users to experiment.

For instance, if I type `zstyle :prompt:pure:success color green` in the terminal, nothing happens. I could type in `prompt pure` to re-init Pure, but it seems like an extra step we should be able to avoid.

Maybe we could re-run this at every preprompt before the colors are checked? Or would that create a lot of overhead?

readme.md
```diff
@@ -93,6 +93,35 @@ prompt pure
...
+|          branch
+|          |       git_arrow
+|          |       |        user|root
+|          |       |        |
+~/dev/pure master* ⇡ zaphod@heartofgold  42s
+venv ❯               |                   |
+|    |               |                   duration
+|    |               git_arrow
+|    success|error
+virtualenv
```
mafredri: This is neat! 😄

readme.md
```diff
@@ -93,6 +93,35 @@ prompt pure
...
+- `branch`, the name of the current branch when in a git repository
+- `error`, the `PURE_PROMPT_SYMBOL` when the previous command *fail*
+- `exec_time`, the execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
+- `git_arrow`, for `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
+- `host`, the hostname when on a remote machine
+- `path`, the current path ie. `PWD`
+- `root`, when the remote user is root
+- `success`, the `PURE_PROMPT_SYMBOL` when the previous command *succed*
+- `user`, the username when on remote machine
+- `virtualenv`, the name of the python virtualenv when in use
```
mafredri: Should we list the default values here as well?

pure.zsh
```diff
@@ -582,6 +583,29 @@ prompt_pure_setup() {
...
+	# most used colors
+	typeset -gA prompt_pure_colors
+	prompt_pure_colors=(
+		branch		242
+		error		red
+		exec_time	yellow
+		git_arrow	cyan
+		host		242
+		path		blue
+		root		white
```
LitoMore:
```suggestion
		root		reset
```

Use `reset` as default color, as https://github.com/sindresorhus/pure/pull/450 mentioned.

mafredri: I think it should be called `default`, I did a quick look through the man-pages and didn't find a `reset` color.

> Most terminals support the colours black, red, green, yellow, blue,  magenta,  cyan  and
white,  which can be set by name.  In addition. default may be used to set the terminal's default foreground colour.

readme.md
bricewge: Yes I think so, at least I see how it can be achieved.

readme.md
bricewge: Thanks.

pure.zsh
bricewge: Changed to `default`.

LitoMore: 👍

pure.zsh
bricewge: The response I got was one `zstyle` lookup is ~4µs so I'll rewrite the PR to only use styles.

mafredri: That sounds affordable 😄, thanks for investigating!

bricewge:
The updated PR add `branch:cached` color. allow live updating of all the colors and show how to 2 state colors at one.
@mafredri @sindresorhus Can I get a second round of review?

pure.zsh
```diff
@@ -103,19 +103,30 @@ prompt_pure_preexec() {
 	export VIRTUAL_ENV_DISABLE_PROMPT=${VIRTUAL_ENV_DISABLE_PROMPT:-12}
 }
 
+# Change the colors if their value are different from the current ones
+prompt_pure_set_colors() {
+	local color_temp
+	for key val in ${(kv)prompt_pure_colors}; do
+		if ! zstyle -T ":prompt:pure:$key" color "$val"; then
+			zstyle -s ":prompt:pure:$key" color color_temp
+			test -n "$color_temp" && prompt_pure_colors[$key]=$color_temp
```
mafredri: I think it would be good to have a way to restore the original color if the zstyle definition is deleted (e.g. `zstyle -d prompt:pure:prompt:\*`).

readme.md
```diff
@@ -93,6 +93,36 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+## Colors
+
+Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default and what part they affect are:
+- `branch` (242), the name of the current branch when in a git repository
+- `branch:cached` (red), the name of the current branch when the data isn't fresh
+- `exec_time` (yellow), the execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
+- `git_arrow` (cyan), for `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
```
mafredri: I see you switched to categories/subcategories, I think it's a good idea. Should we do it for `git` as well? E.g. `git:arrow`.

readme.md
```diff
@@ -102,6 +132,10 @@ autoload -U promptinit; promptinit
 
 # optionally define some options
 PURE_CMD_MAX_EXEC_TIME=10
+# change path color
+zstyle :prompt:pure:path color white
+# prompt color whatever the result of the last command
+zstyle ':prompt:pure:prompt:*' green
```
mafredri: Hmm, I noticed this (star-glob) doesn't work if you also define `:prompt:pure:prompt:success` / `:error`. I'm not sure how `zstyle` decides on the precedence / weight of the settings.

Intuitively, if I have defined `zstyle :prompt:pure:prompt:success color white`, and later do `zstyle :prompt:pure:prompt:'*' color green`, then I think I'd want the glob to override the earlier statement. This might not be feasible / possible though, or maybe not even correct behavior.

If I did it the other way around, maybe I'd want the white to override the green `¯\_(ツ)_/¯`.

Do you have any thoughts on the topic?

readme.md
```diff
@@ -93,6 +93,36 @@ prompt pure
...
+## Colors
+
+Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default and what part they affect are:
+- `branch` (242), the name of the current branch when in a git repository
+- `branch:cached` (red), the name of the current branch when the data isn't fresh
+- `exec_time` (yellow), the execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
+- `git_arrow` (cyan), for `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
+- `host` (242), the hostname when on a remote machine
+- `path` (blue), the current path ie. `PWD`
+- `pompt:error` (red), the `PURE_PROMPT_SYMBOL` when the previous command has *failed*
```
mafredri: => `prompt`.

LitoMore:
Can you fix the merge conflict?

bricewge:
@LitoMore Thanks for the heads-up, I forgot that your PR got merged.

readme.md
```diff
@@ -93,6 +93,39 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+
+## Colors
+
+Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default, and what part they affect are:
```
sindresorhus: Can you link to somewhere that clearly lists all the supported named colors and what the `242` numbered colors refer to?

readme.md
```diff
@@ -93,6 +93,39 @@ prompt pure
...
+- `exec_time` (yellow) - The execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
+- `git:arrow` (cyan) - For `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`
+- `git:branch` (242) - The name of the current branch when in a Git repository
+- `git:branch:cached` (red) - The name of the current branch when the data isn't fresh
+- `host` (242) - The hostname when on a remote machine
+- `path` (blue) - The current path, for example, `PWD`
+- `prompt:error` (red) - The `PURE_PROMPT_SYMBOL` when the previous command has *failed*
+- `prompt:success` (magenta) - The `PURE_PROMPT_SYMBOL` when the previous command has *succeded*
+- `user` (242) - The username when on remote machine
+- `user:root` (default) - The username when the user is root
```
sindresorhus: What does `default` mean? Should be clarified.

readme.md
sindresorhus: Looks great!

sindresorhus:
I think this should be documented https://github.com/sindresorhus/pure/pull/472#discussion_r293116827 as being possible.

readme.md
```diff
@@ -93,6 +93,39 @@ prompt pure
 | **`PURE_GIT_DOWN_ARROW`**        | Defines the git down arrow symbol.                                                             | `⇣`            |
 | **`PURE_GIT_UP_ARROW`**          | Defines the git up arrow symbol.                                                               | `⇡`            |
 
+
+## Colors
+
+Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module) with a pattern of the form `:prompt:pure:$color_name` and style `color`. The color names, their default, and what part they affect are:
+- `exec_time` (yellow) - The execution time of the last command when exceeding `PURE_CMD_MAX_EXEC_TIME`
```
sindresorhus: I would prefer to name this `execution_time`.

I realize you chose `exec_time` because of `PURE_CMD_MAX_EXEC_TIME`, but I would still prefer to make new properties more readable.

readme.md
sindresorhus: Ok, I found this:

> Support for 24-bit true color terminals has been added. Hex triplets can be used when specifying colours for prompts and line editor highlighting. On 88 and 256 colour terminals, a new zsh/nearcolor module allows colours specified with hex triplets to be matched against the nearest available colour. - http://zsh.sourceforge.net/releases.html

We should definitely load `zsh/nearcolor` and document how to specify HEX colors.

mafredri: Oh yeah, investigating `zsh/nearcolor` has been on my todo list, would be great to add support for it. Maybe we also need to add a warning if using hex colors and ZSH < 5.7.

readme.md
ashb: `default` here means the terminal's default colour. Could also be called "uncolored"?

sindresorhus: I think `default` as a keyword is fine. We just need to explain in the docs what it means.

readme.md
bricewge: Support and detection of 24bit color in terminal is a shitshow at the moment, take a look [here](https://gist.github.com/XVilka/8346728) to account for it yourself.
The correct way would be test 24bit support with `echoti RGB`, but it's only supported since the latest ncurses [release](https://www.gnu.org/software/ncurses/), it isn't present in debian stable for example and a lot of application aren't yet supporting it. Testing support with `$COLORTERM`, as written in ZSH's [manual](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fnearcolor-Module), isn't a good idea as it can be first glance as explained [here](https://invisible-island.net/ncurses/ncurses.faq.html#no_colorterm) and [there](https://gist.github.com/XVilka/8346728#gistcomment-2249677).

In my case, I'm using most of the time `tmux` in `alacritty`, using `echoti` would indicate that my terminal can't use 24bit colors, even though I'm currently using them. Which would activate `zsh/nearcolor` and screw up my color palette.
So my stance on this is to let the user load `zsh/nearcolor` by herself, handling the corner case by herself and adding it when the terminal ecosystem will have matured enough on this feature.

readme.md
bricewge: Done.

readme.md
mafredri: I've been thinking about that, and I'm really happy you did the research! From what you've found, it's quite clear to me that we *don't* want to automatically load `zsh/nearcolor`. I see no way around it. All we can do is inform our users, which you've already done in your last commit, muy bien 👌.

sindresorhus: Alright, but we should at least document how to achieve it.

readme.md
```diff
@@ -96,6 +96,11 @@ prompt pure
 
 ## Colors
 
+As explained in ZSH's[manual](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Character-Highlighting) color values can be:
+- a decimal integer up to the maximal number of color your terminal support
+- the name of one of the following nine colours: `black`, `red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `white` and `default` (the terminal’s default foreground)
+- ‘#’ followed by an RGB triplet in hexadecimal format (ex. `#424242`), *only* if your terminal support 24bit colors or when the module [`zsh/nearcolor`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fnearcolor-Module) is loaded
```
sindresorhus: We should show an example of loading `zsh/nearcolor` and using a HEX color in the Example section. I think a lot of users will want this.

readme.md
```diff
@@ -96,6 +96,11 @@ prompt pure
 
 ## Colors
 
+As explained in ZSH's[manual](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Character-Highlighting) color values can be:
+- a decimal integer up to the maximal number of color your terminal support
```
sindresorhus: What does the decimal integer represent? How can users find the number they want? Those are some questions we should answer.

In general, we should not force users to read the horrible ZSH manual to achieve things most user would want.

bricewge: I have explained what it represent but the way to find the which color that I can't explain because it is terminal specific.

Now I understand the reason behind all your comments on the documentation side. And I'm happy to see I'm not the only one thinking that this manual isn't well written! Hell I have even found this gem in the manual:
> [...] the Zsh Development Group accepts *no responsibility* for any brain damage which may occur during the reading [...]

readme.md
bricewge: Done.

readme.md
```diff
@@ -93,6 +93,53 @@ prompt pure
...
+|    prompt
+virtualenv
+```
+
+### RGB colors
+
+There is two way to use RGB colors with the hexadecimal format. The correct way is to use a terminal that support 24bit colors and enable this feature as explained in it's documentation.
+If you can't use such terminal the module [`zsh/nearcolor`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fnearcolor-Module) is here for you, it will map any hexadecimal color to the nearest color in the 88 or 256 color palettes of your termial but without using the first 16 colors since their values can be modifed by the user. Understand that when using this module you won't be able to display true RGB colors, it only allows you to specify colors in a more convenient way. Following is an example on how to use this module:
+```sh
+#.zshrc
```
LitoMore:
```suggestion
# .zshrc
```

bricewge: Changed.

sindresorhus:
I was not able to use 24-bit colors. I have the latest version of ZSH (5.7.1) and iTerm, which both supports 24-bit colors.

This is what I have:

```zsh
zstyle :prompt:pure:path color '#FF69B4'
```

I should have gotten a hot pink path, but it's just black.

Any ideas?

LitoMore:
> I should have gotten a hot pink path, but it's just black.

@sindresorhus It works for me. My ZSH is 5.7.1 and iTerm too. Maybe there is a plugin that has interference factors in your configuration file?

pure.zsh
```diff
@@ -103,19 +103,34 @@ prompt_pure_preexec() {
 	export VIRTUAL_ENV_DISABLE_PROMPT=${VIRTUAL_ENV_DISABLE_PROMPT:-12}
 }
 
+# Change the colors if their value are different from the current ones
+prompt_pure_set_colors() {
+	local color_temp
+	for key val in ${(kv)prompt_pure_colors}; do
```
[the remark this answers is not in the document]
bricewge: C mannerism keep creeping on me. It has been corrected.

ssh352:
when can this feature be merged to master? Thanks!
`zstyle :prompt:pure:path color 230
`

sindresorhus:
This looks good to me now. @bricewge, Thanks for contributing 🙌

sindresorhus:
It's out now: https://twitter.com/sindresorhus/status/1141729038068641797 🎉

dcastil:
Huge thanks for this! Huge win for accessibility for `prompt:success` / `prompt:error`! 😊

jonmorton:
Is zsh 5.2 a new enough version for this change? I started getting `prompt_pure_setup:43: bad set of key/value pairs for associative array'` with this update and pure no longer works.

jonmorton:
I can confirm that reverting this diff fixes pure on zsh 5.2. I think it'd be good to make pure work for this version because it is the current stable one used by CentOS.

mafredri:
@jonmorton 5.2 is supported, this was a regression. Thanks for reporting, it's fixed in 1.10.2. Btw, I even use Pure on 4.3.17 or 5.0.2, occasionally 😛.

Btw, next time please open up a new issue as it helps with tracking 😄.

PR #474, opened 2019-06-04
Add pure10k to the list of ports

romkatv:


sindresorhus:
Can you include a screenshot in the Gist so people can see how it looks like?

romkatv:
Done.

pure10k

romkatv:
> This is not the Gist.

I've included a link to the screenshot in the gist and posted the same link here. If this isn't what you are asking for, could you clarify what I should do?

sindresorhus:
I missed that. It was hidden in the comment. I meant; just add a readme.md file to the gist with the screenshot in it.

romkatv:
Done.

Didn't know gist renders markdown. It looks much better with README.md indeed.

romkatv:
Thanks for merging! And extra thanks for developing Pure. I've learned a lot from it.

PR #477, opened 2019-06-16
Avoid performing prompt reset in zle-line-finish hook

### LINKED ISSUE #476: Terminal Marks don't work
### General information

- Pure version: 1.9.0
- ZSH version: 5.3 (x86_64-apple-darwin18.0)
- Terminal program & version: Terminal Version 2.9.5 (421.2)
- Operating system: macOS Mojave 10.14.5
- ZSH framework: antigen (I get the same problem when not loading antigen too)

I have:
- [] Tested with another terminal program and can reproduce the issue: I did not check other programs

### Problem description
I can't use the go to "previous/next mark" navigation in Termina properly anymore. They should go to the previous/last command prompt. If I load the pure theme now they just go to the first/last line. If I don't load the theme it works properly. I've also tried reinstalling the theme but it didn't work.

### Reproduction steps

1. Load the theme
2. Use the options in Edit > Marks

### My `.zshrc`:

Just loading the theme causes the problem on my machine

```shell
prompt pure
```

If I take out the second line it works again.

mafredri:
We can't perform a prompt reset at this point because it removes the prompt marks inserted by macOS Terminal.

This change will no longer allow the prompt to reset (from ❮ to ❯) when the user hits [Enter] in VI-mode. However, this might actually be the right thing to do.

Fixes #476.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -517,7 +517,8 @@ prompt_pure_reset_vim_prompt_widget() {
 	setopt localoptions noshwordsplit
 	prompt_pure_reset_prompt_symbol
 
-	prompt_pure_reset_prompt
+	# We can't perform a prompt reset at this point because it
+	# removes the prompt marks inserted by macOS Terminal.
 }
 
 prompt_pure_state_setup() {


sindresorhus:
@RexYuan Can you confirm this fixes your problem?

RexYuan:
Sorry for the delay. I just reinstalled with npm but it doesn't seem to work still. Should I have tried something else?

mafredri:
@RexYuan this is not yet part of any release. What you can do is e.g. download [`pure.zsh`](https://raw.githubusercontent.com/sindresorhus/pure/81d8e21914f7213fa018b5d6fb760c9ff35a9081/pure.zsh) from this PR and enable it or replace the one from `npm` with it.

RexYuan:
Thanks @mafredri. I tried the `pure.zsh` you provided and `sync.zsh` from master and indeed this fixed the problem. The marks are working again.

PR #480, opened 2019-06-22
Fixed `exec_time` colour

tonyspiff:


sindresorhus:
There's nothing to fix. It's working fine. Yes, the color has a different name than the environment variable, but we'll eventually fix that when we move all options to zstyle.

PR #481, opened 2019-06-25
Add zstyle color support to PS2

mafredri:
The most important part of this PR changes the PS2 prompt to use the actual `prompt:success` / `prompt:error` colors. The second part adds support for styling the PS2 part that prefixes the prompt indicator (e.g. `... quote`).

Not sure what to call it, went with `prompt2:faded` as it's the faded part of the PS2, but it might as well be called `prompt2:parser_status`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -90,7 +90,7 @@ Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release
 - `path` (blue) - The current path, for example, `PWD`.
 - `prompt:error` (red) - The `PURE_PROMPT_SYMBOL` when the previous command has *failed*.
 - `prompt:success` (magenta) - The `PURE_PROMPT_SYMBOL` when the previous command has *succeded*.
-- `prompt2:faded` (242) - The color for showing the state of the parser in the continuation prompt (PS2).
+- `prompt:continuation` (242) - The color for showing the state of the parser in the continuation prompt (PS2). It's the pink part in [this screenshot](https://user-images.githubusercontent.com/147409/70068574-ebc74800-15f8-11ea-84c0-8b94a4b57ff4.png), it appears in the same spot as `virtualenv`. You could for example matching both colors so that Pure has a uniform look.
 - `user` (242) - The username when on remote machine.
 - `user:root` (default) - The username when the user is root.
 - `virtualenv` (242) - The name of the Python `virtualenv` when in use.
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -110,6 +110,7 @@ Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release
 - `path` (blue) - The current path, for example, `PWD`.
 - `prompt:error` (red) - The `PURE_PROMPT_SYMBOL` when the previous command has *failed*.
 - `prompt:success` (magenta) - The `PURE_PROMPT_SYMBOL` when the previous command has *succeded*.
+- `prompt2:faded` (242) - The color for showing the state of the parser in the continuation prompt (PS2).
```
ashb: How about `prompt:continuation`?

mafredri: The idea here was that we're already signaling it's the continuation prompt by putting it under `prompt2:` (PS2) instead of `prompt:`. The `faded` means it's the faded part _in front_ of the actual prompt indicator (`❯`), we don't want to indicate we're changing the prompt indicator color in any way. It's mostly similar to `virtualenv` in terms of positioning / look, but we can't really piggy-back on that color, unfortunately.

I'm open to alternative naming suggestions, but I'd like to keep the `prompt2:` separation. It's just that in that scenario `continuation` doesn't really make sense. In the future we might also make `prompt4` (debug) colors tweakable and having everything under one `prompt:` could get messy.

sindresorhus:
Why does the color here needs to be customizable? Doesn't seem like something many would care about. What would you personally set it to?

sindresorhus:
@mafredri ⬆️

mafredri:
> Why does the color here needs to be customizable? Doesn't seem like something many would care about. What would you personally set it to?

Sorry, missed this. The main reason I wanted to make this customizable is that `virtualenv` is customizable and being able to change this allows you to create a uniform prompt experience.

sindresorhus:
Can you fix the merge conflict and add the color setting to the ASCII art that shows where in the prompt it is?

readme.md
sindresorhus: Would be useful to include a use-case for wanting to change this color. (Your virtualenv use-case, for example)

mafredri: > Can you fix the merge conflict and add the color setting to the ASCII art that shows where in the prompt it is?

I’ll fix it when I get home today or tomorrow. Regarding the ASCII art, I’m not sure where / how to display it. Adding the continuation prompt there (new line) could look a bit weird. And it’s also not entirely correct to put it on the same line as virtualenv. Perhaps we could add a note of colors “not pictured” in the graph?

sindresorhus: Make the color pink, take a screenshot of your prompt with it and describe that it’s the pink part and link to the screenshot?

mafredri:
@sindresorhus I've amended the feedback and I decided to use @ashb's naming suggestion (thanks!) while I was at it.

PR #491, opened 2019-08-14
Add coloring for git dirty-asterisk

pgoodbread:
I added an option to define the coloring of only the asterisk on the git branch, because I tend to overlook it, if it is the same color

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -123,6 +123,7 @@ prompt_pure_preprompt_render() {
 
 	# Set color for Git branch/dirty status and change color if dirty checking has been delayed.
 	local git_color=$prompt_pure_colors[git:branch]
+	local git_dirty_color=$prompt_pure_colors[git:branch:dirty]
 	[[ -n ${prompt_pure_git_last_dirty_check_timestamp+x} ]] && git_color=$prompt_pure_colors[git:branch:cached]
 
 	# Initialize the preprompt array.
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -106,6 +106,7 @@ Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release
 - `git:arrow` (cyan) - For `PURE_GIT_UP_ARROW` and `PURE_GIT_DOWN_ARROW`.
 - `git:branch` (242) - The name of the current branch when in a Git repository.
 - `git:branch:cached` (red) - The name of the current branch when the data isn't fresh.
+- `git:branch:dirty` (208) - The asterisk showing the branch is dirty.
```
sindresorhus: I think it should be `git:dirty`. While the symbol is shown with the branch, it's not really related to the branch name.

sindresorhus: Can you find a brown color that is a bit less saturated/dimmed?

pgoodbread: Thanks for your feedback! 

I changed my commit, I hope the color fits this time. Otherwise you may suggest one and I can implement it.

pure.zsh
```diff
@@ -672,6 +673,7 @@ prompt_pure_setup() {
 		git:arrow            cyan
 		git:branch           242
 		git:branch:cached    red
+		git:dirty     		 130
```
mafredri: Do we want to default this to a new color? I'd really like this color to default to whatever `git:branch` color is set to. For example, git:branch is 242, this is 242. Or, user changes git:branch to 130, this is 130. That is, unless the user specifies this one explicitly in which case that overrides it.

Is it too hidden / hard to understand? Thoughts @sindresorhus?

PS. Indentation is a bit off here (mixed tab/spaces).

sindresorhus: I don't think we want implicit behavior because `git:branch` color changed. They should all be independent. As for the color change, I think it's nice to make the dirty status a little bit more visible than the branch info. I'm not totally sold on the color though, but we can play around with it later on and decide on what color to use.

sindresorhus:
@pgoodbread Can you fix the merge conflict?

pure.zsh
sindresorhus: We've decided to go with `218` as the default color.

sindresorhus:
Thanks for this awesome contribution :)

PR #493, opened 2019-08-23
Use ellipsis for continuation prompt (PS2)

bricewge:
I had a go at tweaking the PS2 prompt that have found unbalanced. Putting the ellipsis just before the prompt symbol looks more boring and replacing the 3 dots with an UTF-8 character reduce the line width. Related with the discussion at https://github.com/sindresorhus/pure/pull/323.
2019-08-23T17:25:25+02:00
On the left you can see the current style and on the right the style in this PR.

I have no strong opinion on this PR, fell free to ditch it's not to your liking.
WDYT?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -701,8 +701,8 @@ prompt_pure_setup() {
 	# Prompt turns red if the previous command didn't exit with 0.
 	PROMPT+='%(?.%F{$prompt_pure_colors[prompt:success]}.%F{$prompt_pure_colors[prompt:error]})${prompt_pure_state[prompt]}%f '
 
-	# Indicate continuation prompt by ... and use a darker color for it.
-	PROMPT2='%F{242}... %(1_.%_ .%_)%f%(?.%F{magenta}.%F{red})${prompt_pure_state[prompt]}%f '
+	# Indicate continuation prompt by … and use a darker color for it.
+	PROMPT2='%F{242}%_… %f%(?.%F{magenta}.%F{red})${prompt_pure_state[prompt]}%f '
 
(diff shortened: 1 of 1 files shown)


mafredri:
I kinda like this change but it's not clear to me why the order of `...` and `cmdand` was changed.

- Before: `... cmdand ❯`
- Now: `cmdand… ❯`
- Why not? `… cmdand ❯` or `…cmdand ❯`

The idea of the dots was to signify that the previous prompt continues on this line, especially when `cmdand` et co. is missing. When it appears after `cmdand` I don't really understand what it's purpose is.

mafredri:
Hmm, I did give all three variants a try (now/why not), and I do see that "now" kind of looks the best visually, not sure if it conveys the same message but I guess it could work 😄.

Any thoughts on this @sindresorhus?

mafredri:
For reference:
Screen Shot 2019-09-08 at 20 03 42
Screen Shot 2019-09-08 at 20 03 55
Screen Shot 2019-09-08 at 20 04 13

bricewge:
> [...] I do see that "now" kind of looks the best visually

That was my reasoning behind changing the ellipsis position, but it's highly subjective.

> When it appears after cmdand I don't really understand what it's purpose is.

Here it's meaning is to it indicate that the "cmdand" continue on that line and that the user is expected to enter something. I see the ellipsis like the underline dots used on some forms to indicate where to write the content.

mafredri:
Thanks for this @bricewge!

PR #505, opened 2019-10-19
Renice the async worker

symphorien:
Entering a large git repository on a hard drive would noticeably slow down
the whole system. This attempts to solve this problem.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -341,7 +341,7 @@ prompt_pure_async_renice() {
 	fi
 
 	if command -v ionice >/dev/null; then
-		command ionice -c 3 -p $$
+		command ionice --class 3 --pid $$
 	fi
 }
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -337,7 +337,7 @@ prompt_pure_async_renice() {
 	setopt localoptions noshwordsplit
 
 	if command -v renice >/dev/null; then
-		command renice +15 -p $$
+		command renice +15 --pid $$
 	fi
 
 	if command -v ionice >/dev/null; then
--- a/pure.zsh
+++ b/pure.zsh
@@ -331,6 +331,16 @@ prompt_pure_async_git_arrows() {
 	command git rev-list --left-right --count HEAD...@'{u}'
 }
 
+prompt_pure_async_renice() {
+	setopt localoptions noshwordsplit
+	if command -v renice >/dev/null; then
+		command renice +15 -p $$
+	fi
+	if command -v ionice >/dev/null; then
+		command ionice -c 3 -p $$
+	fi
+}
+
 prompt_pure_async_tasks() {
 	setopt localoptions noshwordsplit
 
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -503,6 +514,10 @@ prompt_pure_async_callback() {
 					;;
 			esac
 			;;
+		prompt_pure_async_renice)
+			# we tried to renice the worker, but if it failed, well, there
+			# is nothing we can do.
```
mafredri: This comment should be removed or reworded. It does not necessarily mean the renice failed, it only indicates that the command completed. Status code is located in `$code`, but it doesn't tell the whole story, e.g. `renice` could've failed, but `ionice` succeeded, which would result in "success".

pure.zsh
```diff
@@ -331,6 +331,16 @@ prompt_pure_async_git_arrows() {
 	command git rev-list --left-right --count HEAD...@'{u}'
 }
 
+prompt_pure_async_renice() {
+	setopt localoptions noshwordsplit
+	if command -v renice >/dev/null; then
+		command renice +15 -p $$;
+	fi
+	if command -v ionice >/dev/null; then
+		command ionice -c 3 -p $$;
```
mafredri: Let's remove the trailing semi-colons here (they serve no purpose).

pure.zsh
symphorien: Removed

pure.zsh
symphorien: Done

pure.zsh
```diff
@@ -331,6 +331,16 @@ prompt_pure_async_git_arrows() {
 	command git rev-list --left-right --count HEAD...@'{u}'
 }
 
+prompt_pure_async_renice() {
```
sindresorhus: Can you add a code comment about what this does and why? For posterity.

symphorien: done

pure.zsh
```diff
@@ -331,6 +331,20 @@ prompt_pure_async_git_arrows() {
 	command git rev-list --left-right --count HEAD...@'{u}'
 }
 
+# Try to lower the priority of the worker so that disk heavy operations
+# like `git status` has less impact on the system responsivity.
+prompt_pure_async_renice() {
+	setopt localoptions noshwordsplit
+
+	if command -v renice >/dev/null; then
+		command renice +15 -p $$
```
sindresorhus:
```suggestion
		command renice +15 --pid $$
```

for readability

pure.zsh
```diff
@@ -331,6 +331,20 @@ prompt_pure_async_git_arrows() {
...
+	fi
+
+	if command -v ionice >/dev/null; then
+		command ionice -c 3 -p $$
```
sindresorhus: Same here

pure.zsh
mafredri: These long flags won’t be available e.g. on `busybox` systems, perhaps others. would result in error. The short ones at least are pretty universal.

symphorien: Should I revert to short names, then?

mafredri: I think that would be best. Otherwise I think we will start seeing bug reports about the errors it would cause on such systems.

symphorien: done

PR #510, opened 2019-11-11
Fix 'getting started' instructions

### LINKED ISSUE #507: Manual install steps don't work
### General information

- Pure version: master
- ZSH version: 5.7.1
- Terminal program & version: iTerm
- Operating system: macOS Sierra
- ZSH framework: None

I have:
- [x] Tested with another terminal program and can reproduce the issue: 

### Problem description

The instructions seem to be wrong.

### Reproduction steps

1. Try following the instructions for manual installation, then open a new terminal.
2. You don't see the pure prompt.

### My `.zshrc`:

```shell
prompt pure
```

### Notes

As far as I can tell, the third and final command in the manual installation steps (`fpath+=("$HOME/.zsh/pure")`) only updates the `fpath` variable locally, in the current terminal session. So I tried inserting this line in my `.zshrc` instead (before `autoload...` etc), and this seems to make it work. Not sure if this is the correct approach, but if so, the instructions should be updated to explain that you need to put the `fpath+=` line in your `.zshrc` file.

callumlocke:
Fixes #507.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -43,7 +43,6 @@ That's it. Skip to [Getting started](#getting-started).
 ```sh
 mkdir -p "$HOME/.zsh"
 git clone https://github.com/sindresorhus/pure.git "$HOME/.zsh/pure"
-fpath+=("$HOME/.zsh/pure")
 ```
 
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Bump

PR #517, opened 2019-11-24
Improve issue template and system reporting

mafredri:
Attempted to make the system report part of the issue template and improve formatting.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -603,13 +603,15 @@ prompt_pure_state_setup() {
 prompt_pure_system_report() {
 	setopt localoptions noshwordsplit
 
-	print - "- Zsh: $(zsh --version)"
+	print - "- Zsh: $($SHELL --version) ($SHELL)"
 	print -n - "- Operating system: "
 	case "$(uname -s)" in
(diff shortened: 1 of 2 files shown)


pure.zsh
```diff
@@ -593,7 +593,7 @@ prompt_pure_state_setup() {
 	[[ $UID -eq 0 ]] && username='%F{$prompt_pure_colors[user:root]}%n%f'"$hostname"
 
 	typeset -gA prompt_pure_state
-	prompt_pure_state[version]="1.10.3"
+	prompt_pure_state[version]="1.11.0"
```
sindresorhus: Maybe we can use https://github.com/sindresorhus/replace-in-files-cli to bump this when package.json is bumped.

Example: https://github.com/sindresorhus/modern-normalize/commit/c1d65e3f7daba2b695ccf837d2aef19d586d1ca6#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R19

sindresorhus:
Nice!

pure.zsh
mafredri: I had something similar in mind, that looks perfect!

Issue #502, opened 2019-10-15
quiescent zsh using pure occasionally jumps to 99% cpu usage


skierpage:
### General information

- Pure version: v1.8.0 ?
- ZSH version: 5.7.1 (x86_64-redhat-linux-gnu)
- Terminal program & version: Konsole 19.08.0
- Operating system: Fedora 31 beta KDE spin
- ZSH framework: uhh, none?

I have:
- [ ] Tested with another terminal program and can reproduce the issue: 
- [ ] Followed the [Integration](https://github.com/sindresorhus/pure#integration) instructions for my framework

### Problem description
I run zsh as my default shell, and my terminal is Konsole with three tabs, one where I issue lots of commands and the other two mostly just showing output from programs I run (detached from the shell, they don't show up in `jobs`). I installed pure in September 2018, I think it was a bit after version 1.8.0.  Several times in 2019 I notice in top and htop output that zsh is at 99% CPU. `ps` reveals the zsh at high usage is _not_ in my active shell but one of these inactive ones. This high CPU is not turning the fan on so I'm usually not aware of it, and I'm not sure what triggers this high CPU mode. zsh is not in a git directory. I ran `ltrace` and noticed some
```
strlen("watcher")                                            = 7
strlen("watcher")                                            = 7
strcpy(0x7f2a04d7b690, "watcher")
```
output, and pure uses this string, hence this bug report.

I just updated pure to latest version and instructions (adding it to fpath). After source ~/.zshrc in the pty that was at high CPU, the issue persisted, I will see if this recurs in new shells.

### Reproduction steps

Don't know yet. I've noticed the high CPU usage about 7 times.

I _completely understand_ if you reject this bug report as unreproducible, Thanks for this nice software.

### My `.zshrc`:

```shell
HISTFILE=~/.histfile
HISTSIZE=5000
SAVEHIST=5000
setopt appendhistory notify
bindkey -v

zstyle ':completion:*' completer _complete _ignored
zstyle :compinstall filename '/home/spage/.zshrc'

autoload -Uz compinit
compinit
zstyle ':completion:*' glob 1
zstyle ':completion:*' menu select=long-list select=0
zstyle ':completion:*' list-colors ${(s.:.)LS_COLORS}

limit coredumpsize 2048000

unsetopt hup # don't set HUP to jobs when exiting
bindkey -M viins '' history-incremental-search-backward
bindkey -M vicmd '' history-incremental-search-backward
bindkey -M viins '' beginning-of-line
[...]

PR #518, opened 2019-12-03
Optimize git fetch

### LINKED ISSUE #473: Very inefficient git fetch, pulls all tags and all branches
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/pure/issues/473)

https://github.com/sindresorhus/pure/blob/e7036c43487fedf608a988dde54dd1d4c0d96823/pure.zsh#L297

What this does is fairly inefficient because on large repositories it will fetch the *whole* repository. It will also fetch tags as well, which if added and removed often slow things down a lot.

What I am doing now is editing it so it only pulls a single branch by putting the remote and the current branch as arguments to fetch:
```
refspec=$(git for-each-ref --format='%(upstream)' $(git symbolic-ref -q HEAD) )
if [[ -z $refspec ]]; then
    return
fi
remote=$( printf "%s" "$refspec" | perl -E 'my (undef, undef, $remote, $ref)  = split("/", <>, 4); print $remote')
    ref=$(    printf "%s" "$refspec" | perl -E 'my (undef, undef, $remote, $ref)  = split("/", <>, 4); print $ref' )
fetch_args+=( "$remote" "$ref" )
git -c gc.auto=0 ${fetch_args[*]}
```

There may be a way to get it so we don't have to split the arguments with perl? What my code does is split

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $50.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($50.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/pure/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

mafredri:
Only fetch information for the current branch to minimize the amount of work Pure (well, git) needs to do.

This commit also starts redirecting git stderr to `/dev/null` because in some cases git can start spewing errors like crazy (see #502). We also never use the stderr so there's no point in keeping it around.

Fixes #473.

Future work could include figuring out if we can somehow avoid one or both of the extra subshells (although, they're run in the async job so it will have minor impact).

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -292,6 +292,16 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	local ref
+	ref=$(command git symbolic-ref -q HEAD)
+	local -a remote
+	remote=($(command git for-each-ref --format='%(upstream:remotename) %(refname)' $ref))
+
+	if [[ -z $remote[1] ]]; then
+		# No remote specified for this branch, skip fetch.
+		return 97
+	fi
+
 	# Default return code, which indicates Git fetch failure.
 	local fail_code=99
 
@@ -317,7 +327,13 @@ prompt_pure_async_git_fetch() {
 		fi
 	' CHLD
 
-	command git -c gc.auto=0 fetch >/dev/null &
+	# Only fetch information for the current branch and avoid
+	# fetching tags or submodules to speed up the process.
+	command git -c gc.auto=0 fetch \
+		--quiet \
+		--no-tags \
+		--recurse-submodules=no \
+		$remote &>/dev/null &
 	wait $! || return $fail_code
 
 	unsetopt monitor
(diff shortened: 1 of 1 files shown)


sindresorhus:
With this PR, I noticed that when I have upstream changes (`⇣` showing) and run `git pull --rebase`, the `⇣` doesn't disappear.

mafredri:
@sindresorhus dang, I can't seem to reproduce. Although, I'm using this in combination with the `indestructible-pure` branch. Could the non-disappearance be related to the worker dying?

pure.zsh
```diff
@@ -292,6 +292,16 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	local ref
+	ref=$(command git symbolic-ref -q HEAD)
+	local -a remote
+	remote=($(command git for-each-ref --format='%(upstream:remotename) %(refname)' $ref))
```
romkatv: I'm not an expert in git or Pure but perhaps this would do what you need?

```zsh
local remote="$(command git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>/dev/null)"
```

mafredri: I did consider this, on the one hand, we wouldn't need two git commands (`symbolic-ref` + `for-each-ref`), on the other we avoid parsing the output. With the above command we'd need to split that string on `/` for it to work with `git fetch` (the difference here would be `origin/master` vs `origin master`). And we'd have to consider other aspects like can origin contain `/`? and are we properly handling branches that contain `/`.

But I'd like to understand if these two commands actually produce different output or not (I'm no git expert myself). If the output is the same (ignoring `/` vs space), then it won't at least help with Sindres issue.

romkatv: Good point about parsing ambiguity. I think both branches and remotes can have `/` in them, so there is no way to unambiguously parse the output of my command. If it prints `a/b/c/d`, there are 3 ways to interpret it.

sindresorhus:
> Could the non-disappearance be related to the worker dying?

Could be, yes.

PR #521, opened 2019-12-11
Extend check for Emacs terminals

mxco86:
Fixes display within Emacs for more recent versions where shell use of the EMACS env var has been deprecated. Deprecation notice is here:

- https://github.com/emacs-mirror/emacs/blob/d0e2a341dd9a9a365fd311748df024ecb25b70ec/etc/NEWS.25#L1240-L1244


mafredri:
Thanks!

Issue #509, opened 2019-11-11
Remove empty line


qazip:
There is an empty line between subsequent prompts. This can be seen in the image that it is in the README. For example, there is a line between "Exit 1" and "~".

Is it possible to remove this line? I find it wastes to much space.

PerBothner:
Note that with [shell integration](http://localhost/per/blog/2019/shell-integration-proposal/) a terminal can use some other visual separator between commands, so the empty line is redundant.

0xBADDCAFE:
I locally patch like this:

```diff
diff --git a/pure.zsh b/pure.zsh
index ddb0e30..3dad87d 100644
--- a/pure.zsh
+++ b/pure.zsh
@@ -176,7 +176,7 @@ prompt_pure_preprompt_render() {
 
 	if [[ $1 == precmd ]]; then
 		# Initial newline, for spaciousness.
-		print
+		# print
 	elif [[ $prompt_pure_last_prompt != $expanded_prompt ]]; then
 		# Redraw the prompt.
 		prompt_pure_reset_prompt
```

It is prefer to keep current behavior for other users, so how about create new option value like `PURE_STRIP_NEW_LINE` ?

sindresorhus:
https://github.com/sindresorhus/pure/pull/526#issuecomment-583774870

bishopb:
You can hack around pure's _current_ opinion with a custom implementation of `print`:

```sh
print() { [ 0 -eq $# -a "prompt_pure_precmd" = "${funcstack[-1]}" ] || builtin print "$@"; }
```

This implementation of `print` masks the built-in. When `print` is called with zero arguments inside a function named `prompt_pure_precmd` it does nothing (instead of printing an empty line); otherwise it behaves exactly as the print built-in.

Put this in your .zshrc before initializing pure and you're good to go.

Future updates to how pure behaves may invalidate this hack. Caveat emptor.

james-prado:
@bishopb Is there anyway to prevent the new line **only** when the terminal is `clear`ed or when first opening the terminal?

When I open a new terminal this happens:
![before_clear](https://user-images.githubusercontent.com/13836920/90339076-718af800-dfbc-11ea-8b3d-7603ea9249ec.png)

The custom `print()` function works well, but I only want to run it when the terminal is first opened, or after running `clear`. Do you know how to do something like this?

bishopb:
[...]

PR #526, opened 2020-01-13
Add config option to omit leading new line

flats:
Per https://github.com/sindresorhus/pure/issues/509, this adds an option to omit the new line before the prompt.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -174,7 +174,7 @@ prompt_pure_preprompt_render() {
 	local expanded_prompt
 	expanded_prompt="${(S%%)PROMPT}"
 
-	if [[ $1 == precmd ]]; then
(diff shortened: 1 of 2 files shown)


sindresorhus:
Sorry, not interested in this. Pure is an opinionated prompt. I would recommend Spaceship if you need a lot of customizability: https://github.com/denysdovhan/spaceship-prompt/blob/master/docs/Options.md#prompt ;)

romkatv:
> I would recommend Spaceship if you need a lot of customizability: https://github.com/denysdovhan/spaceship-prompt/blob/master/docs/Options.md#prompt ;)

The easiest alternative would be powerlevel10k given that it can generate exactly the same prompt as Pure. It also has several customization options for Pure style in its configuration wizard, and this includes the option to remove empty lines. See https://github.com/romkatv/powerlevel10k/blob/master/README.md#pure-compatibility.

PR #530, opened 2020-02-13
Move user@host to beginning of prompt

jonathanjsimon:
This makes it far easier to notice that one is root or SSHed because the user@host string is first and it doesn’t move across the screen as one traverses long directory trees.

Before:

Screen Shot 2020-02-12 at 22 31 43

After:

Screen Shot 2020-02-12 at 22 31 22

I can update the readme if these changes are agreeable.


sindresorhus:
Makes sense to me, but I don't really use that feature.

PR #531, opened 2020-02-14
Update zplugin help to zinit

camsteffen:
zplugin has been renamed to zinit. I also added a compile step, which is actually copied from the zinit readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -234,13 +234,13 @@ zplug mafredri/zsh-async, from:github
 zplug sindresorhus/pure, use:pure.zsh, from:github, as:theme
 ```
 
-### [zplugin](https://github.com/zdharma/zplugin)
+### [zinit](https://github.com/zdharma/zinit)
 
 Update your `.zshrc` file with the following two lines (order matters):
 
 ```sh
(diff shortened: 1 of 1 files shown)


romkatv:
This name change from zplugin to zinit is confusing to users. FWIW, in powerlevel10k I'm now listing instructions for both, as if they were separate plugin managers. This way users who think they are using zplugin won't get lost.

sindresorhus:
> This way users who think they are using zplugin won't get lost.

I don't think it's a prompt's job to teach users about a name change. We just show what the current options are.

romkatv:
> > This way users who think they are using zplugin won't get lost.
> 
> I don't think it's a prompt's job to teach users about a name change. We just show what the current options are.

I listed both options to avoid having to teach users about the name change. Listing just `zinit` results in the following exchange:

- Q: How do I use your theme with `zplugin`?
- A: `zplugin` was renamed to `zinit`, please use `zinit` instead.
- Q: I get an error when I try to use `zinit`. What should I do?
- A: You need to upgrade. Here's how...
- Q: Upgrade broke me. What now?
- A: ...

PR #533, opened 2020-02-20
Fix stash symbol showing during git actions

### LINKED ISSUE #532: Bug: Git stash shows during git action
The git stash symbol shows during an interactive rebase or a merge (with conflicts). This happens even when I have not opted into the git stash symbol.

carhartl:
The git stash symbol must never show during a git action, similar to
arrows. It even showed up during an action without opt-in.

Closes #532

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -153,8 +153,8 @@ prompt_pure_preprompt_render() {
 	if [[ -n $prompt_pure_git_arrows ]]; then
 		preprompt_parts+=('%F{$prompt_pure_colors[git:arrow]}${prompt_pure_git_arrows}%f')
 	fi
-	# Git stash symbol (if opted in).
-	if [[ -n $prompt_pure_vcs_info[stash] ]]; then
+	# Git stash symbol (if opted in and not during action).
+	if [[ -n $prompt_pure_vcs_info[stash] && -z $prompt_pure_vcs_info[action] ]]; then
 		preprompt_parts+=('%F{$prompt_pure_colors[git:stash]}${PURE_GIT_STASH_SYMBOL:-≡}%f')
 	fi
 
@@ -270,7 +270,7 @@ prompt_pure_async_vcs_info() {
 	# Export branch (%b), Git toplevel (%R), action (rebase/cherry-pick) (%a),
 	# and stash information via misc (%m).
 	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
-	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a' '%m'
+	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a'
 	if [[ $1 == 0 ]]; then
 		zstyle ':vcs_info:git*+set-message:*' hooks git-stash
 	fi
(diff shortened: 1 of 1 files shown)


camsteffen:
Thanks. Tested the fix and now the stash symbol never shows during a git action.

camsteffen:
Just noticed you asked about this in the bug issue so I responded there.

pure.zsh
```diff
@@ -403,7 +403,6 @@ prompt_pure_async_tasks() {
 		unset prompt_pure_git_dirty
 		unset prompt_pure_git_last_dirty_check_timestamp
 		unset prompt_pure_git_arrows
-		unset prompt_pure_git_stash
```
carhartl: Not related to this fix, this variable was unused, but overlooked and sneaked in with #462. 

I should have at least mentioned that in the commit message, or better yet, have created a separate commit for this cleanup.

pure.zsh
```diff
@@ -153,8 +153,8 @@ prompt_pure_preprompt_render() {
 	if [[ -n $prompt_pure_git_arrows ]]; then
 		preprompt_parts+=('%F{$prompt_pure_colors[git:arrow]}${prompt_pure_git_arrows}%f')
 	fi
-	# Git stash symbol (if opted in).
-	if [[ -n $prompt_pure_vcs_info[stash] ]]; then
+	# Git stash symbol (if opted in and not during action).
+	if [[ -n $prompt_pure_vcs_info[stash] && -z $prompt_pure_vcs_info[action] ]]; then
```
mafredri: The reason the previous implementation didn't work is that when we're rebasing with stash disabled, the contents of `$prompt_pure_vcs_info[stash]` is `no patch applied`.

I haven't delved deeper into the meaning of that but with `%m` (misc) present, some other hooks seem to be running as well.

pure.zsh
```diff
@@ -270,7 +270,7 @@ prompt_pure_async_vcs_info() {
 	# Export branch (%b), Git toplevel (%R), action (rebase/cherry-pick) (%a),
 	# and stash information via misc (%m).
 	zstyle ':vcs_info:git*' formats '%b' '%R' '%a' '%m'
-	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a' '%m'
+	zstyle ':vcs_info:git*' actionformats '%b' '%R' '%a'
```
mafredri: We should probably only modify the formats (and `max-exports`) when stash is actually enabled.

mafredri:
@carhartl since you closed this PR, are you still working on this or not?

sindresorhus:
> What's the purpose of not showing stash during git action, if enabled? To me it doesn't make sense that it disappears but maybe there's some reason behind it?

Focus. You don't really care about any stash during a Git rebase. It will show up again afterwards.

camsteffen:
Maybe I want to stash during an interactive rebase and unstash at a later commit in the rebase. Or stash, start rebase, unstash.

sindresorhus:
I've never needed that. Didn't even know it was a thing, but I guess if there's a real use-case for it, we could keep it.

carhartl:
@mafredi

I closed it because much of it seemed just wrong anyway.

But I’m also not working on it, as there doesn’t seem to be consensus regarding the desired behavior.

camsteffen:
> I've never needed that. Didn't even know it was a thing, but I guess if there's a real use-case for it, we could keep it.

Well yes, the use cases are rare. But I also think it just follows the principle of least surprise. "Hey, where did my stash go! Oh yeah..."

mafredri:
@carhartl alright, thanks for letting us know. @camsteffen opened a PR https://github.com/sindresorhus/pure/pull/535 that could replace this one.

PR #537, opened 2020-03-20
display non-zero exit code

cizmazia:
Would you mind adding an option for displaying non-zero exit codes? I am aware that non-zero exit codes are already being indicated by the red prompt sign, however, that is not very legible. Exit codes are valuable while tinkering compound commands. Thanks!


mafredri:
Would you consider using the right prompt (`RPROMPT`) for that?

For example, type this in your terminal and you'll see what I mean:

```zsh
RPROMPT='%(?..%?)'
```

Otherwise I'd say this is a feature request pending #460 (i.e. a more modular Pure).

cizmazia:
Thanks for your response!

My reason for switching to Pure was that it does not use RPROMT. I will wait for the feature then.

PR #542, opened 2020-04-11
Respect Git config `status.showUntrackedFiles`

jameh:
I use the git config option

```
git config --local status.showUntrackedFiles no
```

in a git repo that tracks files in my `$HOME`, so that `git status` doesn't show me files that I don't care to track. However, I like seeing a dirty branch indicator when there are untracked files in my other git repos, so I do not set the environment variable `PURE_GIT_UNTRACKED_DIRTY`.

These settings result in the dirty indicator in the prompt to always be on when in my `$HOME` dotfiles git repository.

This PR just makes the prompt respect the `status.showUntrackedFiles` setting, and if unset respects the environment variable.

Let me know what you think!

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -283,11 +283,15 @@ prompt_pure_async_vcs_info() {
 prompt_pure_async_git_dirty() {
 	setopt localoptions noshwordsplit
 	local untracked_dirty=$1
+	local untracked_git_mode=$(command git config --get status.showUntrackedFiles)
+	if [[ "$untracked_git_mode" != 'no' ]]; then
+		untracked_git_mode='normal'
+	fi
 
 	if [[ $untracked_dirty = 0 ]]; then
 		command git diff --no-ext-diff --quiet --exit-code
 	else
-		test -z "$(command git --no-optional-locks status --porcelain --ignore-submodules -unormal)"
+		test -z "$(command git --no-optional-locks status --porcelain --ignore-submodules -u${untracked_git_mode})"
 	fi
 
 	return $?


pure.zsh
```diff
@@ -292,11 +292,12 @@ prompt_pure_async_vcs_info() {
 prompt_pure_async_git_dirty() {
 	setopt localoptions noshwordsplit
 	local untracked_dirty=$1
+	local untracked_git_mode=$(command git config --get status.showUntrackedFiles) || 'normal'
 
 	if [[ $untracked_dirty = 0 ]]; then
 		command git diff --no-ext-diff --quiet --exit-code
 	else
-		test -z "$(command git status --porcelain --ignore-submodules -unormal)"
+		test -z "$(command git status --porcelain --ignore-submodules -u${untracked_git_mode})"
```
camsteffen, line 300: Why don't you just remove this flag?

jameh: `-unormal` is much faster than `-uall` because it doesn't recurse into untracked directories, so it's sensible to default to `-unormal` instead of git's default which is `-uall`. For the purposes of a dirty bit, recursing is entirely unnecessary.

camsteffen: Good point - `-uall` should not be allowed.

pure.zsh
```diff
@@ -292,11 +292,12 @@ prompt_pure_async_vcs_info() {
 prompt_pure_async_git_dirty() {
 	setopt localoptions noshwordsplit
 	local untracked_dirty=$1
+	local untracked_git_mode=$(command git config --get status.showUntrackedFiles) || 'normal'
```
camsteffen, line 295: `|| 'normal'` does not work.

You need to validate the output of `git config` since it could be any string that the user puts in their config. I suggest the following:
```sh
untracked_git_mode="$(command git config --get status.showUntrackedFiles)"
if [[ "$untracked_git_mode" != 'no' ]]; then
    untracked_git_mode='normal'
fi
```

jameh: that makes sense! I've force pushed a new commit

sindresorhus:
Can you fix the merge conflict?

sindresorhus:
@camsteffen Bump

sindresorhus:
@jameh Bump

jameh:
I rebased to the current master commit

PR #549, opened 2020-06-11
Eliminate unnecessary Git locking

iloveitaly:
`--no-optional-locks` helps ensure status is a non-blocking operation. Super helpful for large repos. This ensures you can execute git commands while pure is updating git status async.

There's a nice explanation about exactly what this does under the git-status manpage under "Background Refresh"


sindresorhus:
Can you change the minimum Git version in the readme to 2.15.2?

sindresorhus:
Can you change the PR title to describe what problem it solves, not what it adds?

iloveitaly:
@sindresorhus done and done!

PR #553, opened 2020-07-08
Fix '--no-optional-locks' argument position

dedukun:
The `--no-optional-locks` option added in 3a70426 should be placed before `status`, as per the [git status documentation](https://git-scm.com/docs/git-status#_background_refresh), since this is an option of git, and not git-status.


mafredri:
Thanks!

Issue #562, opened 2020-09-04
Feature/Question: Show directory instead of full path


jasonvarga:
Hi! Not a bug, but just a question.

Is it possible to just show the current directory instead of the entire path?

i.e. `myproject` instead of `~/projects/myproject`

mafredri:
This would be solvable by #460, if it gets implemented (but nobody is working on it atm).

Now, if you don't mind making your own changes / forking, it's a simple one line change 😄:

```diff
diff --git pure.zsh pure.zsh
index 2cbd7d0..d8326ae 100644
--- pure.zsh
+++ pure.zsh
@@ -135,7 +135,7 @@ prompt_pure_preprompt_render() {
 	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=($prompt_pure_state[username])
 
 	# Set the path.
-	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%~%f')
+	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%1~%f')
 
 	# Git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
```

jasonvarga:
I don't mind at all. That works beautifully, thank you!

PR #563, opened 2020-09-10
Allow extra data from a _pure_custom_postpath function

marcusirgens:
When using [aws-vault](https://github.com/99designs/aws-vault), it's sometimes very nice to know which profile is active. I've previously used other (slower) prompts, but I really missed having this data in pure.

However, I realise that adding support for a ton of third party stuff is not the goal of this project, so I was wondering if a simple function "plugin" hook could be added? This pull request very simply checks for a function named `_pure_custom_postpath`, and if it exists, it's invoked and the output is appended after the `path` part in the prompt.

As there's a note in the readme saying ...

> Makes an excellent starting point for your own custom prompt.

... I assume that the threshold for adding stuff to pure itself is quite high, so if this is something you explicitly don't want to add I fully understand that. I'm also open to different implementations or ideas for how to potentially make pure "pluggable".

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -137,6 +137,14 @@ prompt_pure_preprompt_render() {
 	# Set the path.
 	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%~%f')
 
+  # Allow extra data from a _pure_custom_postpath function
+	if typeset -f _pure_custom_postpath > /dev/null; then
+		local custom_postpath="$(_pure_custom_postpath)"
+		if [[ -n "$custom_postpath" ]]; then
+			preprompt_parts+=("${custom_postpath}%f")
+		fi
+	fi
+
 	# Git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
 	if [[ -n $prompt_pure_vcs_info[branch] ]]; then


marcusirgens:
I just now realised that #460 deals with this stuff.

mafredri:
@marcusirgens indeed it does, and that would be the ideal solution. But nobody is working on it at the moment AFAIK.

Now, if you don't mind making your own changes / forking, it's a simple one line change 😄:

```diff
diff --git pure.zsh pure.zsh
index 2cbd7d0..d8326ae 100644
--- pure.zsh
+++ pure.zsh
@@ -135,7 +135,7 @@ prompt_pure_preprompt_render() {
 	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=($prompt_pure_state[username])
 
 	# Set the path.
-	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%~%f')
+	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%1~%f')
 
 	# Git branch and dirty status info.
 	typeset -gA prompt_pure_vcs_info
```

marcusirgens:
> ```diff
> 
> diff --git pure.zsh pure.zsh
> 
> index 2cbd7d0..d8326ae 100644
> 
> --- pure.zsh
> 
> +++ pure.zsh
> 
> @@ -135,7 +135,7 @@ prompt_pure_preprompt_render() {
> 
>  	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=($prompt_pure_state[username])
> 
>  
> 
>  	# Set the path.
> 
> -	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%~%f')
> 
> +	preprompt_parts+=('%F{${prompt_pure_colors[path]}}%1~%f')
> 
>  
> 
>  	# Git branch and dirty status info.
> 
>  	typeset -gA prompt_pure_vcs_info
> 
> ```

I'm afraid I don't fully understand what this change does?

mafredri:
Sorry for the confusion @marcusirgens, I thought I was replying to #562 (case of too many tabs 😅)...

---

Back to this PR, as you noticed, #460 is how we'd ideally like to solve this issue as it's more flexible. I'm not entirely opposed to having a possibility for prefix/postfix in Pure, but I'm sure we'd still see requests for changing order or removing some part entirely. A solution which is easy to use and tries to prevent user error (e.g. hide array manipulations) would be preferable here.

sindresorhus:
// @marcusirgens

PR #564, opened 2020-09-17
Show user@host when running in a container

TimSpence:
## Issue
When running a shell inside a docker or LXC container, the pure prompt does not display **user@hostname** as when logged in to a remote computer via SSH.

## Fix
After executing a command on the wrong LXC container yesterday, I found a few reliable methods for determining whether the shell is executing inside a container.
1. In LXC, the environment contains `container=lxc`.
2. In docker and sometimes in LXC, the cgroup for the `init` process indicates "docker" or "lxc".

After the fix, the pure prompt retains the default behavior which hides **user@host** when physically logged in to a computer.

## Testing

LXC:
```
❮ lxc ubuntu dotfile-tester

ubuntu@dotfile-tester ~
❯
```
Ubuntu 20.04 (bare metal):
```
❮ ssh nuc
Welcome to Ubuntu 20.04.1 LTS (GNU/Linux 5.4.0-45-generic x86_64)
...
Last login: Thu Sep 17 18:30:41 2020 from 192.168.1.178

tim@nuc ~
❯
```
OS X Catlina (physical computer, attached with keyboard:
```
❯ hostname
mac-mini

~
❯
```
Docker container:
```
❯ ls
Dockerfile          LICENSE

root@50867f260801 ~
❯
```

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -696,7 +696,7 @@ prompt_pure_state_setup() {
 
 # Return true if executing inside a Docker or LXC container.
 prompt_pure_is_inside_container() {
-	([[ -r /proc/1/cgroup ]] && $(grep -q -E "(lxc|docker)" /proc/1/cgroup )) \
+	([[ -r /proc/1/cgroup ]] && grep -q -E "(lxc|docker)" /proc/1/cgroup ) \
 		|| [[ "$container" == "lxc" ]]
 }
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -677,11 +677,15 @@ prompt_pure_state_setup() {
 	fi
 
 	hostname='%F{$prompt_pure_colors[host]}@%m%f'
-	# Show `username@host` if logged in through SSH.
-	[[ -n $ssh_connection ]] && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
-
-	# Show `username@host` if inside an LXC container.
-	(( $(env | grep -c 'container=lxc') > 0 )) && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
+	# Show `username@host` when:
+	# * logged in through SSH
+	# * inside a docker container
+	# * inside an LXC container
+	if [[ -n $ssh_connection ]] ||
+		( [[ -r /proc/1/cgroup ]] && (( $(grep -cE "(lxc|docker)" /proc/1/cgroup ) > 0 )) ) ||
+		(( $(env | grep -c 'container=lxc') > 0 )); then
+			username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
+	fi
 
 	# Show `username@host` if root, with username in default color.
 	[[ $UID -eq 0 ]] && username='%F{$prompt_pure_colors[user:root]}%n%f'"$hostname"
@@ -790,7 +794,6 @@ prompt_pure_setup() {
 		virtualenv           242
 	)
 	prompt_pure_colors=("${(@kv)prompt_pure_colors_default}")
-
 	add-zsh-hook precmd prompt_pure_precmd
 	add-zsh-hook preexec prompt_pure_preexec
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -677,15 +677,11 @@ prompt_pure_state_setup() {
 	fi
 
 	hostname='%F{$prompt_pure_colors[host]}@%m%f'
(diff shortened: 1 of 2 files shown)


pure.zsh
```diff
@@ -677,8 +677,15 @@ prompt_pure_state_setup() {
...
 
 	hostname='%F{$prompt_pure_colors[host]}@%m%f'
-	# Show `username@host` if logged in through SSH.
-	[[ -n $ssh_connection ]] && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
+	# Show `username@host` when:
+	# * logged in through SSH
+	# * inside a docker container
+	# * inside an LXC container
+	if [[ -n $ssh_connection ]] ||
+		( [[ -r /proc/1/cgroup ]] && (( $(grep -cE "(lxc|docker)" /proc/1/cgroup ) > 0 )) ) ||
```
sindresorhus, line 685: Can you move the container logic to a separate function that's called here? `prompt_pure_is_inside_container` or something.

Use long-flags for `grep`.

pure.zsh
```diff
@@ -677,8 +677,15 @@ prompt_pure_state_setup() {
 	fi
 
 	hostname='%F{$prompt_pure_colors[host]}@%m%f'
-	# Show `username@host` if logged in through SSH.
-	[[ -n $ssh_connection ]] && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
+	# Show `username@host` when:
+	# * logged in through SSH
+	# * inside a docker container
```
sindresorhus, line 682:
```suggestion
	# * inside a Docker container
```

sindresorhus (changes requested):

pure.zsh
TimSpence: Yes, that makes sense.  Updated to use long flags for grep.

pure.zsh
```diff
@@ -691,6 +694,12 @@ prompt_pure_state_setup() {
 	)
 }
 
+# Return true if executing inside a Docker or LXC container.
+prompt_pure_is_inside_container() {
+	( [[ -r /proc/1/cgroup ]] && (( $(grep --count --extended-regexp "(lxc|docker)" /proc/1/cgroup ) > 0 )) ) ||
+		(( $(env | grep --count 'container=lxc') > 0 ))
```
mafredri, line 700:
```suggestion
	([[ -r /proc/1/cgroup ]] && grep -q -E "(lxc|docker)" /proc/1/cgroup) \
		|| env | grep -q 'container=lxc'
```

Proposing a few simplifications and compatibility improvements here:

- `--extended-regexp` is not supported on `busybox` (often used in Docker), `-E` is more compatible
- `-q` keeps grep silent and we can simply rely on it's exit status instead of checking count
- Speed up execution by not using subshells to capture output

mafredri (commented):
Great idea, thanks for the PR!

pure.zsh
mafredri: Another though, can we simply check `container=lxc` vi environment variable instead? Would save us from the execution of env/grep. (Not running LXC myself so can't test.)

TimSpence: Thank you for the suggestions.  I'm having zplug troubles so I will try out a different plugin manager so I can test this efficiently before re-requesting review.

TimSpence: Done.

mafredri: Noticed the grep is still running in a subshell `$( )`, you can remove that still.

TimSpence: Good catch.  I removed it.

pure.zsh
TimSpence: After I changed this to use long options, mafredi [pointed out](https://github.com/sindresorhus/pure/pull/564#pullrequestreview-491333614) that long flags are not supported on busybox.  So I changed this block to use short flags.  AFAIK, `-q` is universally supported.  Can we merge it as it is now, or should I revert back to long flags?

sindresorhus (approved):

mafredri (approved):
LGTM 👍

MattiNieminen:
I personally run a lot of things locally in a container and the container share my hosts hostname and username. Therefore it's not really meaningful for me to show this information. Couple of options:

1) Add an option for disabling this feature.
2) Check that if the username@host is identical between the host and the container, don't show it.

sindresorhus:
> Check that if the username@host is identical between the host and the container, don't show it.

👍

PR #565, opened 2020-09-19
Bump Pure internal version when publishing package

mafredri:
Realized we haven't bumped the internal version in a while, so I wrote a hook for `version`. I haven't tested running this with `np`, only via `npm run version`, but I suspect it should work just fine.


sindresorhus:
Good idea!

PR #569, opened 2020-09-25
Use GIT_OPTIONAL_LOCKS=0 instead of --no-optional-locks

mafredri:
I still access some servers with old versions of git, and #549 / #553 silently breaks Pure functionality when the `--no-optional-locks` option is unavailable.

This gives us backwards-compatibility with old versions of git.

From `man git`:

```
       --no-optional-locks
           Do not perform optional operations that require locks. This
	   is equivalent to setting the GIT_OPTIONAL_LOCKS to 0.
```

**EDIT:** With this we could lower the required `git` version in the Pure readme, for instance `1.9.1` is working just fine.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -291,7 +291,7 @@ prompt_pure_async_git_dirty() {
 	if [[ $untracked_dirty = 0 ]]; then
 		command git diff --no-ext-diff --quiet --exit-code
 	else
-		test -z "$(command git --no-optional-locks status --porcelain --ignore-submodules -u${untracked_git_mode})"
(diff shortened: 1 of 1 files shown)


sindresorhus:
> EDIT: With this we could lower the required git version in the Pure readme, for instance 1.9.1 is working just fine.

I don't see the point of loosening up now. Even the version in the readme is old. The older version we support, the more potential for problems and support needs.

mafredri:
> I don't see the point of loosening up now. Even the version in the readme is old. The older version we support, the more potential for problems and support needs.

Yeah, you're right, I was thinking the wording could be supported/recommended instead of required. But I'm fine with not loosening it up as well. If we can't make something backwards compatible though, we'd (IMO) have to fail early in prompt init via version check or something similar, to avoid silent failure.

PR #578, opened 2020-12-15
Change whether or not is inside container check way to pure zsh

a5ob7r:
It is a little faster(0.001s ~ 0.004s) than before due to no need to call a grep command. Also, no need to consider which version of grep is installed(GNU, BSD, Busybox, or others?).

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -702,7 +702,7 @@ prompt_pure_state_setup() {
 # Return true if executing inside a Docker or LXC container.
 prompt_pure_is_inside_container() {
 	local -r cgroup_file='/proc/1/cgroup'
-	[[ -r "$cgroup_file" && -n "${(M)$(< $cgroup_file):#*(lxc|docker)*}" ]] \
+	[[ -r "$cgroup_file" && "$(< $cgroup_file)" = *(lxc|docker)* ]] \
 		|| [[ "$container" == "lxc" ]]
 }
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -701,7 +701,8 @@ prompt_pure_state_setup() {
 
 # Return true if executing inside a Docker or LXC container.
 prompt_pure_is_inside_container() {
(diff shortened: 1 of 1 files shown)


pure.zsh
```diff
@@ -701,7 +701,8 @@ prompt_pure_state_setup() {
 
 # Return true if executing inside a Docker or LXC container.
 prompt_pure_is_inside_container() {
-	([[ -r /proc/1/cgroup ]] && grep -q -E "(lxc|docker)" /proc/1/cgroup ) \
+	local -r cgroup_file='/proc/1/cgroup'
+	[[ -r "$cgroup_file" && -n "${(M)$(< $cgroup_file):#*(lxc|docker)*}" ]] \
```
mafredri, line 705: This is a good idea. I think this could be made a bit easier to read:

```suggestion
	[[ -r "$cgroup_file" && [[ "$(<$cgroup_file)" = *(lxc|docker)* ]] \
```

Although I can appreciate that the original solution is quite elegant too, but requires deeper knowledge of zsh syntax 😄.

mafredri (approved):

pure.zsh
a5ob7r: Thanks to your review! And your suggestion is certainly true. It is very important to read easily. I will fix it 😄.

By the way, is it typo that the second double open brackets on the above suggested change? I think it will cause a syntax error...🤔

mafredri: Good catch, it was a typo in the sense that I meant to close the previous one (`]] && [[`) 😄. Combining like you did is also fine 👍🏻.

a5ob7r: OK!

a5ob7r:
@mafredri What I need to do for this pull request? Should fix something?

mafredri:
Nah @a5ob7r, it's perfect as it is. I just wanted to get a chance to try it out and then came the holidays 😅. Thanks for your contribution!

PR #581, opened 2021-01-26
Remove antigen and antibody from instruction

dobrovolsky:
Antigen tries to build from the "master" branch, but there isn't a branch with this name.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -219,7 +219,7 @@ Update your `.zshrc` file with the following two lines (order matters). Do not u
 
 ```sh
 antigen bundle mafredri/zsh-async
-antigen bundle sindresorhus/pure
+antigen bundle sindresorhus/pure --branch=main
 ```
 
 ### [antibody](https://github.com/getantibody/antibody)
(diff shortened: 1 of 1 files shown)


mafredri:
Not sure wether or not we should change this here. I think this is worth reporting to the antigen project directly, it should be pulling the default branch otherwise this will become a profound issue for antigen users in the future. It could also be that it's failing because there is already a cached version of the project (with `master` branch), so it's pulling based on that. Not sure what the best solution would be to be honest.

dobrovolsky:
Good point.
Seem like the Antigen project is "dead", so maybe it's better to remove it from readme at all?

sindresorhus:
Yeah. Just remove it from the readme.

sindresorhus (approved):

faizmokhtar:
@dobrovolsky @sindresorhus  Instead of removing can we update the instructions for antigen? I tried `pure` and I need to specify `--branch=main` for the installation to work. 

Something like this:
```
antigen bundle mafredri/zsh-async
antigen bundle sindresorhus/pure --branch=main
```

sindresorhus:
We're not going to document inactive projects. There are enough other ZSH plugin managers.

PR #599, opened 2021-07-04
Add nix shell support

gytis-ivaskevicius:
nix-shell is something similar to python virtual env but more powerful, this patchset adds support for it.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -222,6 +222,13 @@ prompt_pure_precmd() {
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
 
+    # Nix package manager integration.
+    # If this prompt is used from within 'nix shell' - present shell derivation name as
+    # shell name. If derivation name is unset - show 'nix-shell'
+	if [[ -n $IN_NIX_SHELL ]]; then
+		psvar[12]="${name:-nix-shell}"
+	fi
+
 	# Make sure VIM prompt is reset.
 	prompt_pure_reset_prompt_symbol
 
--- a/pure.zsh
+++ b/pure.zsh
@@ -222,9 +222,9 @@ prompt_pure_precmd() {
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
 
-    # Nix package manager integration.
-    # If this prompt is used from within 'nix shell' - present shell derivation name as
-    # shell name. If derivation name is unset - show 'nix-shell'
+	# Nix package manager integration. If used from within 'nix shell' - shell name is shown like so:
+	# ~/Projects/flake-utils-plus master
+	# flake-utils-plus ❯
 	if [[ -n $IN_NIX_SHELL ]]; then
 		psvar[12]="${name:-nix-shell}"
 	fi


pure.zsh
```diff
@@ -222,6 +222,10 @@ prompt_pure_precmd() {
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
 
+	if [[ -n $IN_NIX_SHELL ]]; then
```
sindresorhus, line 225: It would be useful with a succinct code comment explaining what exactly this does and why.

pure.zsh
```diff
@@ -222,6 +222,13 @@ prompt_pure_precmd() {
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
 
+    # Nix package manager integration.
+    # If this prompt is used from within 'nix shell' - present shell derivation name as
+    # shell name. If derivation name is unset - show 'nix-shell'
```
sindresorhus, line 227: Tab-indentation.

It's also not clear to me as a non-nix user, what "shell derivation name" is. Maybe include an example of what the prompt looks like (as text) with "shell derivation name"?

gytis-ivaskevicius: https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/ ;)

`Nix` package manager has a concept of `derivations` which basically is a compilation unit that results in a file or directory (Can be package/script/config/etc). And derivation `name` is a user-defined value that can be anything. In the case of `mkShell` derivation builder most likely you'd set `name` argument according to your project name.

Can you please tell me what kind of comment is expected here and ill more or less copy-paste it since I am really not sure what else am I supposed to type. I feel like to explain it to non-nix users I would straight up need to copy-paste few pages of documentation, it is a very different build tool/package manager, with very different ideas compared to most. Getting a hang of it [takes some time](https://raw.githubusercontent.com/gytis-ivaskevicius/high-quality-nix-content/master/memes/nix-learning-curve.png)

sindresorhus: My pure prompt looks like:

```
~/dev
❯
```

I'm looking for a code comment with the above just how it looks like with the Nix stuff.

gytis-ivaskevicius: Here is an example of how it looks:
image

And here is ASCII copy-paste:
```
~/Projects/flake-utils-plus github-actions*
flake-utils-plus ❯ unset name

~/Projects/flake-utils-plus github-actions*
nix-shell ❯
```

Would you like me to drop this ASCII as a comment above that if statement?

sindresorhus: Yes

sindresorhus (approved):

pure.zsh
```diff
@@ -222,6 +222,13 @@ prompt_pure_precmd() {
 		export VIRTUAL_ENV_DISABLE_PROMPT=12
 	fi
 
+	# Nix package manager integration. If used from within 'nix shell' - shell name is shown like so:
+	# ~/Projects/flake-utils-plus master
+	# flake-utils-plus ❯
+	if [[ -n $IN_NIX_SHELL ]]; then
+		psvar[12]="${name:-nix-shell}"
```
mafredri, line 229: Should we consider compatibility with e.g. virtualenvs? Can there be a virtualenv active simultaneously with a nix shell? Should both be visible?

mafredri (commented):
I think this looks alright and would probably work in 99% of cases, only potential issue I see is that the nix shell may override an active virtualenv (if it's possible for them to be active at the same time).

pure.zsh
gytis-ivaskevicius: Yeah, I thought about it and decided to implement it in this manner since showing both venv and nix-shell would not make much sense.

And since nix-shell may contain venv and not another way around - it is implemented in a manner so it would overwrite venv name

mafredri (approved):

andresilva:
Would be nice to have a way to disable this.

gytis-ivaskevicius:
`unset IN_NIX_SHELL`?

andresilva:
`IN_NIX_SHELL` wasn't set by this code, it works but you're assuming it isn't used for anything else besides `pure`. I think having a `NIX_SHELL_DISABLE_PROMPT` makes sense, just like it's done for `VIRTUAL_ENV` support.

mafredri:
> `IN_NIX_SHELL` wasn't set by this code, it works but you're assuming it isn't used for anything else besides `pure`. I think having a `NIX_SHELL_DISABLE_PROMPT` makes sense, just like it's done for `VIRTUAL_ENV` support.

We don't have anything like this for virtualenvs 🤔, we simply respect the user option of `VIRTUAL_ENV_DISABLE_PROMPT` if it is set, it's part of regular virtualenvs though, not Pure.

Is there any similar mechanism for nix shell that we could piggy-back on? We'd like to avoid adding a new configuration option for this.

gytis-ivaskevicius:
No, there is no such logic to piggyback on. 
@andresilva Can you elaborate on your usecase? Whatever your use case it sounds pretty odd :thinking:

andresilva:
> We don't have anything like this for virtualenvs 🤔, we simply respect the user option of `VIRTUAL_ENV_DISABLE_PROMPT` if it is set, it's part of regular virtualenvs though, not Pure.

I don't use virtualenv so I didn't know this, I assumed it was something specific to pure.

> @andresilva Can you elaborate on your usecase? Whatever your use case it sounds pretty odd 🤔

My use case is quite simple: I don't like this and I want to disable it. IMO the proposed solution to unset a variable that wasn't set by pure is unsatisfactory. If I have some other script that relies on `IN_NIX_SHELL` for detection then it will be broken.

mafredri:
> IMO the proposed solution to unset a variable that wasn't set by pure is unsatisfactory. If I have some other script that relies on IN_NIX_SHELL for detection then it will be broken.

I 100% agree. I've not been personally able to evaluate how nix shell can/is used since this is pretty much the first I've heard of it. Obviously if it's something that's enabled 24/7, then displaying the (excess) information is not ideal.

Would it make senses to never display `nix-shell` (i.e. `$name` is unset), instead only display it when `$name` has a value? My thought process is that it could be valuable to know when you're in a different "environment", or whatever it is.

If we were to add a configuration option for this, perhaps it could be added as a `zstyle` (`zstyle :prompt:pure:environment:nix-shell show yes`)?

andresilva:
> Would it make senses to never display `nix-shell` (i.e. `$name` is unset), instead only display it when `$name` has a value? My thought process is that it could be valuable to know when you're in a different "environment", or whatever it is.

I don't want to bore you with the details but I don't use `nix-shell` directly but instead use [lorri](https://github.com/nix-community/lorri), this means that I'll always have a `$name` set even though I'm not interested in seeing it.

> If we were to add a configuration option for this, perhaps it could be added as a `zstyle` (`zstyle :prompt:pure:environment:nix-shell show yes`)?

That would work for me.

PR #602, opened 2021-07-27
Fix version bumping script

LitoMore:
~~Let the script `version` know when it's time to run.~~

Make sure `version` changes are added to staged.

I tried `preversion`, it still requires us to add changes to staged.


sindresorhus:
`version` is already a npm script though. So this should work, but I have noticed it doesn't commit the version change, so maybe we need to use `preversion` or something. I didn't have a chance to check yet.

LitoMore:
That's because we don't a process can trigger the `vesion` script. We need a tool like the [husky](https://github.com/typicode/husky) (trigger the script when `pre-commit`) or just use the `prepublishOnly` to trigger the script before `npm publish`.

sindresorhus:
`$ npm version` triggers the `version` script. `np` uses `npm version`.

https://docs.npmjs.com/cli/v7/commands/npm-version

LitoMore:
I checked the [source code](https://github.com/npm/cli/blob/latest/lib/version.js). I didn't see any logic to trigger the `npm run version` when we run `npm version`.

I believe they are different. `npm version` is a built-in command for output the version info. But the `version` in the `scripts` field is just a script.

LitoMore:
Oh, I see. 

> I have noticed it doesn't commit the version change.

I think we need to add our changes to staged (`git add`).

PR #604, opened 2021-08-15
Do not show hostname when running in GitHub Codespaces

MarshallOfSound:
When running in GitHub Codespaces pure thinks it's running inside a container (correctly) but we don't need the codespaces hostname when using the integrated terminal.

Tested this change using my dotfiles and the default codespaces image.


sindresorhus (approved):

mafredri (approved):

mafredri:
Thanks for the PR!

PR #613, opened 2021-10-08
Allow configuring nix-shell integration using zstyle

andresilva:
As discussed here: https://github.com/sindresorhus/pure/pull/599#issuecomment-882473718


andresilva:
If it's not too much of an ask it would be nice to get a new release if/when this is merged ☺️ (and I will take care of updating the nix package).

mafredri:
@andresilva done, thanks for you contribution!

PR #616, opened 2021-11-23
Add suspended job indicator

e-roux:
Hello,

this brings a symbol ✦ in the prompt in case of suspended job in the background.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -131,6 +131,11 @@ prompt_pure_preprompt_render() {
 	# Initialize the preprompt array.
 	local -a preprompt_parts
 
+	# Suspended jobs in background.
+	if ((${(M)#jobstates:#suspended:*} != 0)); then
+		preprompt_parts+='%F{$prompt_pure_colors[suspended_jobs]}✦'
+	fi
+
 	# Username and machine, if applicable.
 	[[ -n $prompt_pure_state[username] ]] && preprompt_parts+=($prompt_pure_state[username])
 
@@ -175,11 +180,7 @@ prompt_pure_preprompt_render() {
 		$cleaned_ps1
 	)
 
-    PROMPT=""
-    if ((${(M)#jobstates:#suspended:*} != 0)); then
-      PROMPT+='%F{$prompt_pure_colors[git:dirty]}✦ '
-    fi
-    PROMPT+="${(j..)ps1}"
+	PROMPT="${(j..)ps1}"
 
 	# Expand the prompt for future comparision.
 	local expanded_prompt
@@ -816,6 +817,7 @@ prompt_pure_setup() {
 		prompt:error         red
 		prompt:success       magenta
 		prompt:continuation  242
+		suspended_jobs       red
 		user                 242
 		user:root            default
 		virtualenv           242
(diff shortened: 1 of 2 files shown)


pure.zsh
```diff
@@ -175,7 +175,11 @@ prompt_pure_preprompt_render() {
 		$cleaned_ps1
 	)
 
-	PROMPT="${(j..)ps1}"
+    PROMPT=""
```
sindresorhus, line 178: Use tab-indentation.

pure.zsh
```diff
@@ -175,7 +175,11 @@ prompt_pure_preprompt_render() {
...
+    if ((${(M)#jobstates:#suspended:*} != 0)); then
+      PROMPT+='%F{$prompt_pure_colors[git:dirty]}✦ '
```
sindresorhus, line 180: This need a new color instead of reusing one.

pure.zsh
```diff
@@ -175,7 +175,11 @@ prompt_pure_preprompt_render() {
 		$cleaned_ps1
 	)
 
-	PROMPT="${(j..)ps1}"
+    PROMPT=""
+    if ((${(M)#jobstates:#suspended:*} != 0)); then
```
sindresorhus, line 179: Needs a code comment.

sindresorhus (changes requested):

pure.zsh
mafredri: This should be moved higher up so that it's part of the `preprompt_parts` array (instead of doing `PROMPT` manipulation here).

mafredri (approved):
Looks good to me, although I'm not entirely convinced of the position of it (I'd prefer after the path).

But I think it's fine to merge as-is, we can introduce an option to disable and/or move it in case it gets in the way.

sindresorhus:
I think it should be after the path.

sindresorhus:
You also need to add it to https://github.com/sindresorhus/pure#colors

e-roux:
After the path, it is common to have the git information. I fear the region will be overloaded with symbols, especially when on git dirty status therefor I tend to have it at the very beginning.

readme.md
```diff
@@ -156,6 +156,7 @@ Colors can be changed by using [`zstyle`](http://zsh.sourceforge.net/Doc/Release
 - `prompt:error` (red) - The `PURE_PROMPT_SYMBOL` when the previous command has *failed*.
 - `prompt:success` (magenta) - The `PURE_PROMPT_SYMBOL` when the previous command has *succeeded*.
 - `prompt:continuation` (242) - The color for showing the state of the parser in the continuation prompt (PS2). It's the pink part in [this screenshot](https://user-images.githubusercontent.com/147409/70068574-ebc74800-15f8-11ea-84c0-8b94a4b57ff4.png), it appears in the same spot as `virtualenv`. You could for example matching both colors so that Pure has a uniform look.
+- `suspended_jobs` (red) - The four pointed star showing jobs runs in the background.
```
sindresorhus, line 159:
```suggestion
- `suspended_jobs` (red) - The `✦` symbol indicates that jobs are running in the background.
```

Issue #584, opened 2021-02-09
Cannot add pure to prompt themes, connected to MB M1?


christophrumpel:
### General information

I'm using the pure theme now for two years without any issues. Yesterday I switched to a new MacBook Air M1 and now I'm having some issues. I cannot set the prompt to pure with `prompt pure`.

#### System report (output of `prompt_pure_system_report`):

zsh: command not found: prompt_pure_system_report

#### I have:

- [x] Tested with another terminal program and can reproduce the issue: iTerm
- [x] Followed the [integration](https://github.com/sindresorhus/pure#integration) instructions for my framework

### Problem description

When I try to add the theme with `prompt pure`:

Result:
```
Usage: prompt <options>
Options:
    -c              Show currently selected theme and parameters
    -l              List currently available prompt themes
    -p [<themes>]   Preview given themes (defaults to all)
    -h [<theme>]    Display help (for given theme)
    -s <theme>      Set and save theme
    <theme>         Switch to new theme immediately (changes not saved)

Use prompt -h <theme> for help on specific themes.
```

When I show the list of available themes with `prompt -l`, `pure` is `not` in this list.
I have installed it, like the years before, with `npm install --global pure-prompt`.

I'm not sure what the problem is. Could it be connected to the M1 MacBook?

### Reproduction steps

1. MacBook Air M1
2. Install pure through npm global
3. Try to set the theme with the code below from my `.zshrc` file.

### My `.zshrc`:

```shell
# Activate Oh-My-Zsh
source $ZSH/oh-my-zsh.sh

# Change Pure Colors
zmodload zsh/nearcolor
zstyle :prompt:pure:git:branch color green

# Activate Pure Theme
# Make sure to remove this if you use another theme
autoload -U promptinit; promptinit
prompt  pure
```

mafredri:
What is the output of the npm install? If you're not using zsh from homebrew then automatic linking won't work and there's additional steps to follow, which should've been shown in the npm install output (i.e. adding pure to `$fpath`).

christophrumpel:
Hey and thanks @mafredri ,

the output is:

```shell
npm install --global pure-prompt  


changed 1 package, and audited 2 packages in 1s

found 0 vulnerabilities
```

Nothing shown there. It is just strange that it worked the last two years

`If you're not using zsh from homebrew then automatic linking won't work...`

Isn't oh-my-zsh using zsh? Anything else I can try?

mafredri:
[...]

PR #621, opened 2022-02-04
Improve getting started for M1 mac users

teru01:
resolve
https://github.com/sindresorhus/pure/issues/584

reference
https://github.com/sindresorhus/pure/issues/584#issuecomment-989054653

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -104,6 +104,7 @@ Initialize the prompt system (if not so already) and choose `pure`:
 
 ```sh
 # .zshrc
+fpath+=/opt/homebrew/share/zsh/site-functions # necessary if you use M1 mac
 autoload -U promptinit; promptinit
 prompt pure
 ```


mafredri:
Hmm, this should not be required normally.

My guess is that for this to happen you may have installed Zsh via Homebrew (`/opt/homebrew/bin/zsh`), but are using the system (`/bin/zsh`) instead.

We should perhaps add a check during install instead.

BeyondEvil:
I don't know what to tell you, dude.

I'm setting up a brand new M1 mac, I have not installed zsh via brew. I **did** however install pure using brew.

```sh
% ls -la /opt/hombrew/bin | grep zsh
%
% which zsh
/bin/zsh
% whence -p zsh
/bin/zsh
```

Before adding the `fpath+=...` the prompt did not load. Instead it printed `Usage` for the prompt command. 🤷‍♂️ 

I'm happy to provide more information and help troubleshoot if you want.

mafredri:
@BeyondEvil the Homebrew installation method is not maintained by us, and it looks like it depends on zsh so it would've been pulled in when you installed it via Homebrew.

https://formulae.brew.sh/formula/pure#default

MrCrunchwrap:
I am using the system `zsh` and I also had to add the line for pure to work

likun7981:
I wrote a pure installation script for M1  https://github.com/likun7981/install-pure

mafredri:
@likun7981 I hope you don't mind but I marked your comment off-topic since it is an install script for a **fork** of Pure.

likun7981:
> @likun7981 I hope you don't mind but I marked your comment off-topic since it is an install script for a **fork** of Pure.

Ok I’m sorry, I delete it right now

PR #623, opened 2022-02-17
Add check for diamond character when inside a toolbox

gbrlgian:
Two simples lines to check if user is currently inside a [Toolbox](https://github.com/containers/toolbox) and add the diamond character from the default Toolbox prompt in place of the user name, just before the path.
toolbox


pure.zsh
```diff
@@ -702,6 +702,9 @@ prompt_pure_state_setup() {
 	# Show `username@host` if inside a container and not in GitHub Codespaces.
 	[[ -z "${CODESPACES}" ]] && prompt_pure_is_inside_container && username='%F{$prompt_pure_colors[user]}%n%f'"$hostname"
 
+        # Show `⬢` if inside a toolbox.
+        [[ "$HOSTNAME" = "toolbox" ]] && [[ $(which toolbox) = "/usr/bin/toolbox" ]] && username='%F{$prompt_pure_colors[host]}⬢%f'
```
mafredri, line 706: We use tabs for indentation.

Would it not also make sense to have the `prompt_pure_is_inside_container` check pass since Toolbox is container based?

mafredri (changes requested):
Before we go any further with this PR, I'd like to understand how ⬢ is indicative of Toolbox? I'm totally unfamiliar with it myself.

And what does it currently look like when you enter Toolbox, pre-patch?

mafredri:
Also, this seems like it kind of steps on virtualenv territory, and as such I think it would make more sense (out of a consistency perspective) to show ⬢ before the prompt symbol.

gbrlgian:
> Also, this seems like it kind of steps on virtualenv territory, and as such I think it would make more sense (out of a consistency perspective) to show ⬢ before the prompt symbol.

I chose that placement, right at the beginning, to be familiar to the average default-prompt Toolbox user, but I can amend another commit placing it where you specified, if needed!

Default prompt for your consideration:
https://blog.christophersmart.com/2020/04/11/easy-containers-on-fedora-with-toolbox/

pure.zsh
gbrlgian: I think Toolbox wouldn't pass that tes, since it's not based on Docker/LXC.

example

gbrlgian: Putting this check makes the diamond not render anymore. I suspect it is caused by the patchy way Toolbox works.
example

gbrlgian: Inserting that check makes it not work anymore. I suspect it is caused by the patchy way Toolbox works.
example

PR #624, opened 2022-04-21
Monitor the value of environment variables

benozol:
pure is really nice, thanks!

this PR allows to monitor the value of environment variables with pure. I found it quite useful when working with commands that use environment variables.

image


benozol:
trying to upstream my config – and getting a much better solution… thanks!

PR #632, opened 2022-06-08
Add a reference to Starship-Pure in the ports section of readme

charitarthchugh:
I have created a config for Starship, a cross shell prompt, that tries to mimic Pure. If any configuration settings needs to be updated, I would be happy to do so to match Pure. 

EDIT:
There is an official preset that mimics Pure more closely in the Starship Docs. The only downside in my opinion to this is that it also limits language support to languages that pure supports.


charitarthchugh:
Thank you @sindresorhus !

sindresorhus:
?

charitarthchugh:
I thought you accepted the pr so just wanted to say thank you

PR #635, opened 2022-06-14
Fixes ssh_connection comparison

aanzolaavila:
Checking inside `prompt_pure_state_setup` takes quite a long time to execute, I'm adding an option to check or not this.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -664,9 +664,10 @@ prompt_pure_state_setup() {
 	setopt localoptions noshwordsplit
 
 	# Check SSH_CONNECTION and the current state.
+	local check_ssh=${true:-$PURE_CHECK_SSH_CONNECTION}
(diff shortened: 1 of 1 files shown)


mafredri:
Would you mind running this on the slow machine (execution times will suffice, no need for all output):

```shell
time who -m
time who
```

And the output from this command:

```shell
ls -lha /var/run/?tmp
```

I'd like to figure out why these commands are running so slowly on your machine. Does it use spinning HDDs?

sindresorhus:
I would prefer fixing the underlying issue than adding more options.

aanzolaavila:
Hi, this is the output from my slow computer
Screen Shot 2022-06-15 at 11 43 45 AM

sindresorhus:
Closing as it should be properly fixed. We cannot add options for everything. I suggest moving this to an issue instead.

PR #638, opened 2022-07-08
Fix README about installation via Homebrew 

Udomomo:
This PR fixes README for installation via Homebrew.  
It's a small fix, but due to this I overlooked this part, where the actual solution for my issue is clearly stated.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -83,7 +83,7 @@ That's it. Skip to [Getting started](#getting-started).
 brew install pure
 ```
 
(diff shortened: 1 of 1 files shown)


mafredri:
The _not_ is supposed to be there. When using Zsh from Homebrew, modifying `fpath` should not be necessary. But when using Zsh that ships with macOS (`/bin/zsh`), it needs to be added.

Care to share more about your setup that caused the need to modify `fpath`? For instance, it's possible to have Homebrew Zsh installed whilst using the macOS one, in which case the hint is there to update the `fpath`.

Issue #614, opened 2021-10-11
Option to chanhe user@hostname in prompt 


acantepie:
I want to change user@hostname on prompt , is it possible to doing that ?

sindresorhus:
Change it how exactly?

reportaman:
Is it possible to always show username@hostname, even when I have not ssh'ed into a machine?

sindresorhus:
@reportaman Anything is possible by modifying the prompt yourself, but it's not something we plan to support built-in.

mafredri:
@sindresorhus what's the action-point for this issue? I'm not entirely sure if it's adding an option for disabling the showing of `@hostname`, or to detect `hostname == hostname` between host and container (as per https://github.com/sindresorhus/pure/pull/564#issuecomment-774739621)?

Of the two, I think only the former is possible.

sindresorhus:
@mafredri The former.

Tr0nJ4volta:
I would second @poetaman's ask for an option to always show the user@host in the prompt. I am happy to open up a new issue and/or submit a PR.

leejkee:
> [@sindresorhus](https://github.com/sindresorhus) Is it possible to always show username@hostname, even when I have not ssh'ed into a machine?

Just add this line to your .zshrc file:
```shell
prompt_pure_state[username]='%F{$prompt_pure_colors[user]}%n%f%F{$prompt_pure_colors[host]}@%m%f'
```

Issue #488, opened 2019-07-31
Feature Request: Dictionary to nickname hostname shown in prompt


milezhacks:
Suppose that powers-beyond-your-control set your hostname to something like `b12ed7860b3a`

So your prompt looks like:
~ user@b12ed7860b3a
>

Could you add a away to, instead of using the hostname auto-variable, use a host nickname dictionary to translate the actual hostname to a user's preference.

For example (in pseudo-code):
```
   # Read in aliases (HostDict)
   source ~/.host-alias

   if $HOSTNAME in HostDict:
      NICEHOST = HostDict[$HOSTNAME]
      HOSTPROMPT=$NICEHOST
   else
      HOSTPROMPT=%H
   fi
   PROMPT=%u@$HOSTPROMPT
```

ELLIOTTCABLE:
Implemented this in #641.

sindresorhus:
This is not something I plan to add. However, you can already do this by overriding the `HOST` variable in your `~/.zshrc` on the remote machine:

```sh
case $HOST in
    b12ed7860b3a) HOST='docker-dev' ;;
esac
```

This makes `%m` (which Pure uses for the hostname) expand to the alias. It also affects the terminal title and anything else reading `$HOST`, which is usually what you want.

Issue #136, opened 2015-07-03
Server hostname while on SSH


aspyatkin:
Hello, I don't know if it's an issue, but I have encountered the following:
<img width="794" alt="screenshot" src="https://cloud.githubusercontent.com/assets/4273812/8502070/db1c479c-21bc-11e5-87eb-045724b32270.png">
I am connecting to my server with installed oh-my-zsh and pure, and it seems I don't get the full server hostname (aspyatkin.com) in prompt, only part of it (aspyatkin).

mafredri:
Hi, this is because `pure` is actually displaying the short hostname, `hostname -s`.

arthurvr:
Is this something we plan to change? I actually like that it displays the short version.

sindresorhus:
No. Happy to receive doc improvements if this is unclear though.

PR #641, opened 2022-07-26
Add configuration option for mapping username/hostname per-host

ELLIOTTCABLE:
This PR implements a configuration-option that allows users to provide a mapping of replacement user- and host-names for systems they do not control, but often SSH into.

As an example, we use `nspawn` at my organization, and I do not have control over my username — this PR allows me to replace the infinitely-annoying `elliott.cable@spawnbox-febox3-elliottcable` with `ec@nspawn3`:

```sh
typeset -gA PURE_HOST_MAP
PURE_HOST_MAP[spawnbox-febox3-elliottcable]="elliott.cable:ec@nspawn3"
```

As for the rest, well, this is best understood from the [documentation I've added to the README](https://github.com/ELLIOTTCABLE/pure/tree/userhost-map#hostnames).

(Implementing this in a sufficiently-general way ended up being mildly messy; enough so that I'll understand if you don't wish to upstream this improvement. If not, I intend to main this feature in my fork; could you add `ELLIOTTCABLE/pure` to the list of maintained forks?)

This implements #488, and solves #614 and #136.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -118,6 +118,40 @@ prompt_pure_set_colors() {
 	done
 }
 
+# This uses %m (hostname) and %n (username) evaluated at setup-time; and if they're found in a user-
+# configured map, replaces them (for the rest of the shell-session) with specified alternative(s).
+prompt_pure_produce_userhost_to_var() {
+	local uservar=$1 hostvar=$2 rawhost rawuser
+
+	# Fallback to using the traditional prompt-replacement sequences
+	typeset -g "$uservar"="%n"
+	typeset -g "$hostvar"="%m"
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for the pull request. However, I don't think this is something we want to maintain. What we could consider is exposing some kind of minimal hook to let you implement this yourself.

ELLIOTTCABLE:
> However, I don't think this is something we want to maintain.
good eyes, and a reassuring response; I like slim.

any ideas of the best approach? i'm limited on time, but can take a brief glance if it's approachable to an outsider.

PR #645, opened 2022-08-10
Make sure GPG-agent does not mess up the tty

### LINKED ISSUE #366: pinentry-ncurses in async process mangles inputs for the rest of the session
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/pure/issues/366)

### General information

- Pure version: git (2017-11-13)
- ZSH version: 5.4.2
- Terminal program & version:
  - putty
  - conemu -> wsl
  - alacritty (etc)
- Operating system: NixOS
- ZSH framework: Toasty Zsh

I have:
- [x] Tested with another terminal program and can reproduce the issue.
- [x] Followed the [Integration](https://github.com/sindresorhus/pure#integration) instructions for my framework (none exist but I can write them if required).

### Problem description

If pure ends up invoking pinentry-ncurses (e.g through ssh-agent) in an async task (e.g checking `git@...` repository's status), input becomes mangled on both sides.

### Reproduction steps
(note: not tested with a "normal" ssh-agent, nor outside nixos)

1. SSH into your system (for it to be autodetected).
2. Start gpg-agent with ssh-agent emulation, and default pinentry (ssh sessions are autodetected, and use pinentry-ncurses). (without your key currently being cached/open)
3. Enter an ssh+git repository (`git@...`).

Expected behavior:
pinentry-ncurses pops up, you enter the password and continue on your merry way

Actual behavior:
pinentry-ncurses pops up (in the async process), and mangles inputs (I haven't looked into it very deeply, but it seems like half your inputs go to asynced-pinentry and half go into the shell).

### My `.zshrc`:

```shell
prompt pure
export GPG_TTY=$(tty)
export SSH_AUTH_SOCK=$(gpgconf --list-dirs agent-ssh-socket)
gpg-connect-agent updatestartuptty /bye > /dev/null
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $210.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)
- [chipsenkbeil chipsenkbeil](https://issuehunt.io/u/chipsenkbeil) ($100.00)
- [antoineco antoineco](https://issuehunt.io/u/antoineco) ($50.00)

### Submitted pull Requests
- [#645 Make sure GPG-agent does not mess up the tty](https://issuehunt.io/r/sindresorhus/pure/pull/645)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/pure/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

sheeviZi:
When gpg-agent is used as an ssh-agent and configured with pinentry-ncurses,
it messes up the terminal immediately after you cd into a git repository with an ssh remote.

Fixes #366.

I can help you reproduce it if you ask.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#366: pinentry-ncurses in async process mangles inputs for the rest of the session](https://issuehunt.io/repos/5890857/issues/366)
---
</details>

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -329,6 +329,9 @@ prompt_pure_async_git_fetch() {
 	# solves issue #366
 	export GPG_TTY=$TTY
 
+	# tell the child process to not expect any input from this zpty
+	exec /dev/null > $TTY
+
 	local -a remote
 	if ((only_upstream)); then
 		local ref
--- a/pure.zsh
+++ b/pure.zsh
@@ -324,6 +324,11 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	# if gpg-agent is setup to handle ssh keys for git fetch; then
+	# make sure it uses this zpty instead of corrupting the parent tty
+	# solves issue #366
+	export GPG_TTY=$TTY
+
 	local -a remote
 	if ((only_upstream)); then
 		local ref
--- a/pure.zsh
+++ b/pure.zsh
@@ -324,12 +324,11 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
-	# if gpg-agent is setup to handle ssh keys for git fetch; then
-	# make sure it uses this zpty instead of corrupting the parent tty
-	# solves issue #366
+	# If gpg-agent is set to handle SSH keys for `git fetch`, make
+	# sure it uses this zpty instead of corrupting the parent TTY.
 	export GPG_TTY=$TTY
 
-	# tell the child process to not expect any input from this zpty
+	# Tell the child process to not expect any input from this zpty.
 	exec /dev/null > $TTY
 
 	local -a remote


pure.zsh
```diff
@@ -324,6 +324,13 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	# if gpg-agent is setup to handle ssh keys for git fetch; then
+	# make sure it uses this zpty instead of corrupting the parent tty
+	# solves issue #366
+	if export -p GPG_TTY &>/dev/null && [[ "$GPG_TTY" != "$(tty)" ]] ; then
```
mafredri, line 330: Should we check `[[ -n $GPG_TTY ]]` instead or is there special meaning if the user has done `export GPG_TTY=`? Alternatively we could do `(( $+GPG_TTY ))` which is a pretty common pattern in Zsh to check if a variable is set. We could also replace `$(tty)` with `$TTY` to avoid creating a subshell.

Or even more radical, is there a reason we shouldn't always do `export GPG_TTY=$TTY`? I can think of one: the user users a dedicated terminal for GPG auths, but I don't know if that's possible or happens in practice. But in that case the above TTY swap would also break that workflow.

sheeviZi: If the variable is not exported then `gpg-connect-agent` would not see it. But I accept your suggestion since the nuance isn't really important here.

I will implement your suggestions.

sheeviZi: done

antoineco:
@sheeviZi that looks very simple, but what is the intended effect of this change? After applying this patch, I don't have the problem of the mangled input anymore (hurray! 🏆), but I also don't see any pinentry prompt. This is perfectly fine with me, I'm just trying to understand what's going on under the hood.

pure.zsh
```diff
@@ -324,6 +324,11 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	# if gpg-agent is setup to handle ssh keys for git fetch; then
+	# make sure it uses this zpty instead of corrupting the parent tty
+	# solves issue #366
+	export GPG_TTY=$TTY
```
antoineco, line 330: This unfortunately doesn't solve the problem.

The tty in the context of this function's execution is different from the user's current tty, so the same problem occurs (pinentry on a different tty, mangled input).

There is a trivial way to verify it: add `touch /tmp/mytty && echo $TTY >> /tmp/mytty` before this line, and check the contents of `/tmp/mytty` after changing directory to a Git repository.
E.g. when my tty is `/dev/ttys000`, the tty seen by pure's async functions is `/dev/ttys001`.

antoineco (changes requested):

pure.zsh
sheeviZi: The tty in this context is meant to be different, because pure lauches this function `prompt_pure_async_git_fetch` asynchroneoulsy in a different tty than the main one.

If the issue still occurs then possibly you didn't setup the ssh config in `~/.ssh/config` to always update the gpg agent with the current tty like mentioned in the archlinux wiki [Configure pinentry to use the correct TTY
](https://wiki.archlinux.org/title/GnuPG#Configure_pinentry_to_use_the_correct_TTY).

antoineco: You're right, I didn't have `Match host * exec "gpg-connect-agent UPDATESTARTUPTTY /bye"` inside my SSH config. Thanks for the tip!

That seems to eliminate the issue indeed. I still see a pinentry process owned by gpg-agent, but it seems like your second commit is supposed to fix it. I'll try it separately and report.

antoineco (commented):
That works for me ✅  The prompt is being suppressed _and_ the pinentry process seems to exit properly in the background.

pure.zsh
antoineco: I would suggest mentioning this command or referencing the page you linked somewhere, because this config is external to Pure, yet critical to be able to use Pure with such SSH setup.

I can volunteer to update the wiki with a section dedicated to GPG-related issues such as this one.

sheeviZi:
@antoineco Thanks for your time.

antoineco:
@sheeviZi a note about the bounty on this issue.

If you want to claim the reward, you will have to
1. Log in at https://issuehunt.io/r/sindresorhus/pure/issues/366 using your GitHub account
2. Click "Submit a Pull Request" on the right hand side of the page
3. Paste the URL of this PR

pure.zsh
sheeviZi: I guess by wiki you mean the [pure wiki](https://github.com/sindresorhus/pure/wiki/)? Should it be about how to setup gpg-agent as a replacement for ssh-agent or should it be just a note about the updating the tty in ssh config?

antoineco: Yes the Pure wiki. I would just add a note about the necessity to add `Match host * exec ...` to the SSH config, because this might be quite unconventional (unless you are used to having daemons using SSH in the background). At least for me, updating gpg-agent's TTY in `.zshrc` was always enough, until I ran into this issue with Pure.

sheeviZi: Can you please review this: https://github.com/sindresorhus/pure/wiki/Advise-on-using-%60gpg-agent%60-as-an-alternative-%60ssh-agent%60. Feel free to edit as you see fit.

antoineco:
@sheeviZi I saw the request for review but I am only providing feedback as a user, not a maintainer. You will have to wait for a maintainer to get their eyes on this in order to have it merged (if accepted as a solution to the problem).

sheeviZi:
@antoineco  Are you ok with the [wiki page](https://github.com/sindresorhus/pure/wiki/Advise-on-using-%60gpg-agent%60-as-an-alternative-%60ssh-agent%60)?

antoineco:
Yes I think that looks good, and it's useful information paired with the fix from this PR.

sheeviZi:
@mafredri is this ok for merge?

sheeviZi:
@sindresorhus can you please take a look in order to close the linked issue; Otherwise I am not able to claim the bounty on issuehunt.

pure.zsh
```diff
@@ -324,6 +324,13 @@ prompt_pure_async_git_fetch() {
 	# Set SSH `BachMode` to disable all interactive SSH password prompting.
 	export GIT_SSH_COMMAND="${GIT_SSH_COMMAND:-"ssh"} -o BatchMode=yes"
 
+	# If gpg-agent is set to handle SSH keys for `git fetch`, make
+	# sure it uses this zpty instead of corrupting the parent TTY.
+	export GPG_TTY=$TTY
+
+	# Tell the child process to not expect any input from this zpty.
+	exec /dev/null > $TTY
```
mafredri, line 332: This broke git fetch for me. I've never seen this syntax for closing stdin, feels like it would close both stdin and stdout/break the TTY? Looks to me like this pretty much stop execution in it's tracks.

If we want to close just stdin, that'd be best accomplished by:

```sh
exec <& -
```

Should we revert the entire PR or just remove this `exec` line @sindresorhus? I'm not sure if my above suggestion will fix the issue that was attempted to fix, and don't have the facility to test it.

sindresorhus: I removed the `exec` part: https://github.com/sindresorhus/pure/releases/tag/v1.20.3

// @sheeviZi

sheeviZi: Fixed in #647 with an alternative solution. Tested locally and made sure that `pinentry-ncurses` closes and while git fetch stays unaffected.

sheeviZi:
Thank you @sindresorhus for merging this PR. Just mention me with @sheeviZi if you need anything in the future.

PR #646, opened 2022-08-17
fix: readme.md default values

frahman944:


mafredri:
@sindresorhus should we revert this one? Technically there is no default value on these, although we default to `1` if they're not set. So essentially default value is `1`, not `0`, and the only legit configuration option for these is to set them to either `1` (the default) or `0`. So I don't see a problem with the original docs.

sindresorhus:
@mafredri Even I got confused by the syntax. How about we just change the docs to have default value of `1`?

PR #647, opened 2022-09-11
Set an empty GPG_TTY for git fetch with gpg-agent

sheeviZi:
This is a signal to gpg-agent to not expect user interaction.

Followup to #645.

Fixes the issue reported in https://github.com/sindresorhus/pure/pull/645#discussion_r967800932.


sindresorhus (approved):

mafredri (approved):
Looks good to me :+1:.

PR #652, opened 2023-01-05
Update zsh-async to v1.8.6

mafredri:
Noticed it's been a while since async was updated, and there are some important compatibility fixes (e.g. when users alias cat).

### DIFF
--- a/async.zsh
+++ b/async.zsh
@@ -3,12 +3,12 @@
 #
 # zsh-async
 #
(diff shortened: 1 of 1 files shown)


sindresorhus:
Did you test this update with Pure to ensure it didn't break anything?

mafredri:
Yeah I've been using it for >1 year and so have the main branch users. 😄

PR #654, opened 2023-02-13
Create .gitignore for *.zwc

migimigi:
Created a .gitignore file that ignores the zsh word code (*.zwc) compiled with the zcompile command for zsh scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# zsh word code files
+*.zwc


sindresorhus:
Why should this be here? Pure does not call `zcompile`.

migimigi:
This was not a problem with Pure, but with prezto. I mistakenly reported it in the wrong place. I will close this issue. Thank you!

PR #655, opened 2023-03-20
Detect when inside an OCI container

blissd:
Add detection of OCI containers so pure can detect when it is running inside of podman.


sindresorhus (approved):

mafredri (approved):
Looks good, thanks!

PR #690, opened 2024-12-03
Support non-Git VCS systems via external functions

stek29:
this is an example on how non-standard VCS systems can be supported in pure, with ability to provide different level of support by defining custom hooks.

I haven't implemented `_alias` hook though.


sindresorhus:
I appreciate the pull request, but this is not something we are likely to merge. Pure optimizes for the common case. We don't intend to add options for everything. Check out https://starship.rs instead.

PR #693, opened 2025-04-15
Support k8s detection

cih9088:
Detect it by checking if a k8s token file exists.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -727,10 +727,12 @@ prompt_pure_is_inside_container() {
 	local -r cgroup_file='/proc/1/cgroup'
 	local -r nspawn_file='/run/host/container-manager'
 	[[ -r "$cgroup_file" && "$(< $cgroup_file)" = *(lxc|docker)* ]] \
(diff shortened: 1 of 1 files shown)


sindresorhus:
The `containerd` addition looks good.

I'm not sure the token file check is needed though. K8s pods are containers running via containerd/Docker/CRI-O, so the cgroup checks should already detect them. Is there a specific case where they aren't detected?

sindresorhus:
I'd prefer to keep this simpler rather than adopting the full systemd approach. That's too much complexity for a prompt theme.

Looking at the systemd code, the most valuable additions are just two generic marker file checks:
- `/.dockerenv` (Docker)
- `/run/.containerenv` (Podman)

Could you update this PR to:
1. Keep the `containerd` addition ✓
2. Add the two marker file checks above
3. Remove the k8s-specific token check

Most k8s environments should be detected via `$container=oci` (set by containerd/CRI-O) or `/.dockerenv` (Docker-based k8s). If there are edge cases where that doesn't work, users can set the `$container` env var in their pod specs.

This gives us better coverage while staying generic and minimal.

PR #704, opened 2026-02-03
add PURE_GIT_EXE for custom git command configuration

SherAndrei:
Allows users to specify alternative git commands or wrappers while maintaining backward compatibility.

Particularly useful for WSL2 where accessing Windows filesystem using WSL2 programs is incredibly slow, see https://github.com/microsoft/WSL/issues/4401.

Default remains `command git` for alias bypassing; set PURE_GIT_EXE to override (e.g., `export PURE_GIT_EXE='git'` for git alias, or `export PURE_GIT_EXE='hub'` for GitHub wrapper).

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -253,12 +253,20 @@ prompt_pure_precmd() {
 	fi
 }
 
+prompt_pure_git_exe() {
+	if [[ -n "${PURE_GIT_EXE}" ]]; then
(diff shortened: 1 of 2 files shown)


SherAndrei:
Thanks for the feedback! I see. Will it help with debugging to print such env string in `prompt_pure_system_report`?

PR #708, opened 2026-02-10
Fix SSH user@host color regression in psvar prompt construction

### LINKED ISSUE #707: SSH user@host segment can become invisible after (#706)
After commit 987ae71 (Use psvar for prompt construction), the SSH user@host segment in Pure’s preprompt can render in a color that is effectively invisible on certain themes (confirmed with Ghostty + Gruvbox Dark and Light), making it appear missing unless selected/highlighted.

**Expected Behavior:**
`user@host` should render with configured Pure colors and remain visible across dark themes.

**Actual Behavior:**
`user@host` is rendered with a very dark/black color and can be invisible against the background.

**Cause:**
`prompt_pure_state[username]` contains nested prompt color expressions and without an extra evaluation pass, nested colors are not resolved correctly.

I'm making a PR with a quick fix for this, but I haven't had time to look into any other affected code paths. Probably worth verifying.

tansanrao:
Ensure nested color variables in the username/host segment are evaluated when building PROMPT so SSH user@host remains visible on dark themes after the psvar refactor.

Fixes #707

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -862,7 +862,7 @@ prompt_pure_setup() {
 	#
 	# Preprompt line: each %(NV..) section only renders when its psvar is non-empty.
 	PROMPT='%(12V.%F{$prompt_pure_colors[suspended_jobs]}%12v%f .)'
-	PROMPT+='%(13V.${prompt_pure_state[username]} .)'
+	PROMPT+='%(13V.${(e)prompt_pure_state[username]} .)'
 	PROMPT+='%F{${prompt_pure_colors[path]}}%~%f'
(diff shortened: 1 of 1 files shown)


mafredri (commented):
Thanks for catching this and for the PR.

While this is technically safe considering the input, an important note here:

If the username state variable ever would end up containing `$(echo injection)`. Before psvar refactor this would _not_ have been executed. But with this change, it would.

We could consider moving the username construction to also use the psvar system rather than conditionally building the state var. But that’s a slightly larger change.

PR #718, opened 2026-03-28
make current directory configurable as psvar[21]

### LINKED ISSUE #660: Feature Request: Truncate prompt path
Expose function to "hot truncate" and one to reset.

Will be useful for those deep paths. Would be great to pass argument "diralias" and have that as a temporary home.
So if I traverse further, the new folders get tacked on.

Then one to reset of course.

My apologies if this is already possible by some other means, couldn't find a way myself!

Love the prompt, appreciate your work.

willnorris:
If psvar[21] is set, use that value as the current directory, defaulting to the existing `%~`.  This allows, for example, truncating the directory based on available column width.

Fixes #660 (insomuch as it makes it easy for users to customize this)

---

I have long used a personal fork of pure to add support for making the prompt adapt to both available window height and width.  With the recent psvar changes, this requires only two very minimal changes, this being the first.   This approach seemed like the least opinionated (since people want different things for this directory) and also the least invasive.  If you're curious, you can see an example of [how I'm using this here](https://github.com/willnorris/dotfiles/blob/main/config/zsh/startup/prompt.zsh#L59).

I recognize this will also conflict with #717 which uses psvar[21] for nodejs version.  I can change this to 22 if that PR lands first.

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -841,7 +841,7 @@ prompt_pure_setup() {
 	typeset -g prompt_pure_git_branch_color=$prompt_pure_colors[git:branch]
 
 	# Construct PROMPT once, both preprompt and prompt line. Kept
-	# dynamic via variables and psvar[12-20], updated each render
+	# dynamic via variables and psvar[12-21], updated each render
 	# in prompt_pure_preprompt_render. Numbering starts at 12 for
 	# legacy reasons (Pure originally used psvar[12] for virtualenv)
 	# and to avoid collisions with low psvar indices which users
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the PR! Unfortunately, this is not something I plan to merge.

- I closed the underlying issue (#660) as path truncation isn't a good fit for Pure. The full path is one of the most valuable things a prompt shows, and zsh's built-in named directories (`hash -d`) are a better solution for long paths.
- For more advanced customization like this, the recently added `prompt_pure_precustom` hook (`psvar[22]`/`psvar[23]`) provides a general-purpose escape hatch without needing a dedicated psvar slot.

PR #720, opened 2026-05-09
Add customizable preprompt segments

### LINKED ISSUE #460: Add support for modifying the pre-prompt (add / edit / remove pre-prompt parts)

apples-kksk:
Fixes #460

## Changes
- Split the first prompt line into named preprompt segments while keeping the default rendered PROMPT unchanged.
- Add helpers to add, replace, remove, and position preprompt segments without maintaining a fork.
- Document the new segment API and default segment order.

## Verification
- `zsh -n pure.zsh`
- `zsh -n async.zsh`
- Non-interactive zsh check covering add/set/remove/before/after segment behavior
- `git diff --check`

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#460: Add support for modifying the pre-prompt (add / edit / remove pre-prompt parts)](https://oss.issuehunt.io/repos/5890857/issues/460)
---
</details>

### DIFF
--- a/pure.zsh
+++ b/pure.zsh
@@ -118,6 +118,121 @@ prompt_pure_set_colors() {
 	done
 }
 
+prompt_pure_preprompt_segment_index() {
+	setopt localoptions noshwordsplit
+
+	local name=$1 segment
+	integer index=1
+	for segment in $prompt_pure_preprompt_order; do
+		if [[ $segment == $name ]]; then
+			REPLY=$index
+			return 0
+		fi
+		(( index++ ))
+	done
+
+	REPLY=0
+	return 1
+}
+
+prompt_pure_rebuild_prompt() {
+	setopt localoptions noshwordsplit
+
+	local segment
+	typeset -g PROMPT=
+	for segment in $prompt_pure_preprompt_order; do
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks, but I'm not interested in AI generated PRs. I can prompt an AI too.

